PR #1, opened 2026-05-31
fix: Fix RP2350B registry for large package, thx @davidbuzz

emolitor:
Merge fix from @davidbuzz fixing RP2350 registry for the larger RP2350B package

### DIFF
--- a/os/hal/ports/RP/RP2350/rp_registry.h
+++ b/os/hal/ports/RP/RP2350/rp_registry.h
@@ -111,15 +111,26 @@
 #define RP_HAS_PWM11                        TRUE
 
 /* ADC attributes.
-   Note: RP2350A (QFN-60) has 5 channels (4 external + temp on ch4)
-         RP2350B (QFN-80) has 9 channels (8 external + temp on ch8)
-   Default configuration is for RP2350A (Pico 2). */
+   RP2350A (QFN-60): 4 external inputs on GPIO26-29 (AINSEL 0-3),
+                     temp on AINSEL 4.
+   RP2350B (QFN-80): 8 external inputs on GPIO40-47 (AINSEL 0-7),
+                     temp on AINSEL 8.
+   Select QFN-80 variant by defining RP2350B_QFN80 in the board build
+   flags. */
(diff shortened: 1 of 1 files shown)


gdisirio (approved):
Approved — clean, well-scoped change for the RP2350B (QFN-80) ADC variant; channel counts, temp channel, and base pins match the datasheet. Non-blocking note: RP_ADC_RROBIN_MASK stays 0x01FF0000 for both variants — fine if that's the register field width, just worth a confirm. Leaving the merge to you to test the flow.

emolitor:
The RP2040 uses a 5-bit mask 0x001F0000, while RP2350 uses a 9-bit mask 0x01FF0000 for both variants. Even though the pins are not bonded on the RP2350A the MUX still exists on the silicon and the ADC will return random floating values, effectively the same as a floating pin. Its up to the caller to either set a MASK per their hardware configuration or ensure they only use determinate (non-floating) values.

davidbuzz:
Hi

chibios-sheriff (commented):

**Scope:** header-only change to the RP2350 ADC registry; adds the RP2350B (QFN-80) variant behind `RP2350B_QFN80`, targets `main`. Clean refactor (common attrs pulled out of the conditional).

**Correctness:** channel counts / temp channel / base pins match the parts — A: 5 / ch4 / GPIO26–29, B: 9 / ch8 / GPIO40–47; `AINSEL_BITS=4` still covers 9 channels. Looks sound.

**Questions (non-blocking):**
- `RP_ADC_RROBIN_MASK` stays `0x01FF0000` (9 channels) for *both* variants — intended as the register field width, or should RP2350A use a narrower 5-channel mask? Please confirm.
- Confirm ADC driver code sized by `RP_ADC_NUM_CHANNELS` handles 9 on the B path (no fixed-size-5 arrays).

**Mechanical gates** (`stylecheck.py` + an RP2350 build) to be confirmed separately — not run in this pass.

**Verdict:** advisory-LGTM pending the `RROBIN_MASK` confirmation. Human maintainer decides the merge.

PR #2, opened 2026-06-02
Adding dev container

Ellerbach:
This PR adds a dev container. Making it easier to build any target withing a container without having to add anything on the machine.
This has been tested building few targets.
In case something is seen missing over time, this can be adjusted.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu-22.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    gcc-arm-none-eabi \
+    libnewlib-arm-none-eabi \
+    gdb-multiarch \
+    openocd \
+    build-essential \
+    make \
+    gawk \
+    git \
+    python3 \
+    python3-pip \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN ln -sf /usr/bin/gdb-multiarch /usr/local/bin/arm-none-eabi-gdb
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     gawk \
     git \
     python3 \
+    python-is-python3 \
     python3-pip \
     && rm -rf /var/lib/apt/lists/*
 
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,5 @@
-FROM mcr.microsoft.com/devcontainers/base:ubuntu-22.04
+# this will always pull the latest version of the base image, which is updated regularly by Microsoft
+FROM mcr.microsoft.com/devcontainers/base:ubuntu
 
 ENV LANG=C.UTF-8 \
     LC_ALL=C.UTF-8 \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,10 @@
 FROM mcr.microsoft.com/devcontainers/base:ubuntu-22.04
 
+ENV LANG=C.UTF-8 \
+    LC_ALL=C.UTF-8 \
+    LANGUAGE=C.UTF-8 \
+    PYTHONUTF8=1
+
 RUN apt-get update && apt-get install -y --no-install-recommends \
     gcc-arm-none-eabi \
     libnewlib-arm-none-eabi \
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -1,0 +1,30 @@
+# Dev Container for ChibiOS
+
+This folder contains the Visual Studio Code Dev Containers configuration for this repository.
+
+## Files
+
+- `devcontainer.json`: VS Code container definition (image build, extensions, post-create hook).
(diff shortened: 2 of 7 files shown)


Ellerbach:
Note: do not merge yet :-) still addressing one element in the dev container :-)

Edit: works now properly and exactly as I want!
I had to bump the PyYAML to 6.0.2 as previous versions won't properly install in recent versions of Python.

chibios-sheriff (commented):

**Scope:** tooling-only — adds `.devcontainer/` (VS Code Dev Container), `.editorconfig`, a readme tree note, and bumps PyYAML 5.4.1→6.0.2. Targets `main`, additive and isolated; no OS/HAL code touched.

**Should-fix:**
- `post-create.sh` does not initialize submodules. This repo uses submodules (`ext`, `tools/ftl`), so a fresh container would have them empty and most builds needing `ext` (FatFS/lwIP/…) would fail. Please add `git submodule update --init --recursive` to `post-create.sh`.

**Notes (non-blocking):**
- Ubuntu 22.04's `gcc-arm-none-eabi` is older (~10.x) than the canonical ChibiStudio toolchain (14.3.1) — fine for general builds, just be aware of version drift.
- PyYAML 5.4.1→6.0.2 is a sensible accompanying fix (5.4.1 fails to build on modern Python; 6.x also drops the old unsafe-load surface).
- `.editorconfig` applies `[*]` repo-wide — harmless but broad.

**Mechanical gates:** no `.c/.h` changed (stylecheck N/A); the image wasn't `docker build`-tested in this pass.

**Verdict:** advisory-LGTM once submodule init is added to `post-create.sh`. Human maintainer decides the merge.

Ellerbach:
> Ubuntu 22.04's gcc-arm-none-eabi is older (~10.x) than the canonical ChibiStudio toolchain (14.3.1) — fine for general builds, just be aware of version drift.

For that one, it's not that straight forward to put in the dev container. I can look at this later in a further iteration if that become a clocker!

emolitor:
> > Ubuntu 22.04's gcc-arm-none-eabi is older (~10.x) than the canonical ChibiStudio toolchain (14.3.1) — fine for general builds, just be aware of version drift.
> 
> For that one, it's not that straight forward to put in the dev container. I can look at this later in a further iteration if that become a clocker!

Why not use 26.04 as the baseline container which should get you GCC 15.2 IIRC?

chibios-sheriff (commented):
🤖 **chibios-sheriff — re-review (advisory)**

**Scope:** dev container; re-review after new commits addressing the prior findings. Targets `main`.

**Blocking:** none.
**Should-fix:** ✅ resolved — `post-create.sh` now runs `git submodule update --init --recursive`, so a fresh container populates `ext`/`tools/ftl` before any build.
**Notes:** ✅ the Ubuntu `gcc-arm-none-eabi` (~10.x) vs canonical 14.3.1 caveat is now documented in the script too. PyYAML bump remains sensible.

**Mechanical gates:** no `.c/.h` changed (stylecheck N/A); image not `docker build`-tested this pass.

**Verdict:** advisory-LGTM — both prior points addressed. Human maintainer decides the merge.

Ellerbach:
> Why not use 26.04 as the baseline container which should get you GCC 15.2 IIRC?

'lll use it without a tag, we'll always get the latest. Makes things easy to start with (rebuilding to make sure all works as expected and will push the change)

chibios-sheriff (commented):
🤖 **chibios-sheriff — re-review (advisory)**

**Scope:** dev container; re-review after `5d2d30e5` (base-image bump + pip flag + line-ending fix). Targets `main`.

**Blocking:** none.

**Should-fix:**
- The toolchain NOTE in `post-create.sh` (lines 16–17) is now **stale** — it still reads *"Ubuntu 22.04 ships gcc-arm-none-eabi ~10.x"*, but the base image is now `ubuntu` (floating latest, currently 24.04 → ~13.x), so it conveys the wrong version. Suggest generalizing: drop the `22.04`/`~10.x` specifics and keep the point that the Ubuntu-packaged `gcc-arm-none-eabi` may differ from the canonical ChibiStudio 14.3.1.

**Notes (non-blocking):**
- **`--break-system-packages` is correct and in fact necessary** here: Ubuntu 24.04+ (PEP 668) marks the system Python externally-managed, so the `pip install` would otherwise fail. Good follow-through from the base bump. (A venv would avoid touching system Python, but it is fine for an ephemeral dev container.)
- **Floating base `:ubuntu` (untagged)** is a deliberate non-reproducibility tradeoff — a container built later can pull a newer toolchain. Fine for a convenience dev container, and you documented it. If reproducibility ever matters, pin a dated tag.
- **Line endings now LF-only** ✅ — verified; the `.sh` will not break in-container.

**Mechanical gates:** no `.c/.h` changed (stylecheck N/A); CI does not `docker build` the container, so the image was not exercised this pass.

**Verdict:** **comment** — effectively mergeable; the only open item is the stale toolchain comment (cosmetic but misleading). Human maintainer decides the merge.

gdisirio:
Merging via admin bypass: this branch was created before the `Review Mechanical Checks` workflow existed on `main`, so the workflow isn't on the head branch and the three required checks can't run here (they'd be trivially green anyway — no `.c/.h` changed). Reviewed and advisory-LGTM; the only open item is the stale toolchain comment, which we'll tidy on `main`. Future PRs branch from current `main` and run the checks normally.

PR #4, opened 2026-06-02
Adding minimal Git Action, labeler and PR template

Ellerbach:
Adding minimal Git Action, labeler and PR template

- have codestyle and other things like this in the action
- have a simple PR template
- labeler for easier identification of what is changed

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,234 @@
+# ---------------------------------------------------------------------------
+# Automatic PR label configuration for actions/labeler v5.
+#
+# Label rules use the "any" (match if any glob matches) strategy by default.
+# Each top-level key is the exact label name that must exist in the repo.
+# Create the labels once via the GitHub "Labels" settings page (or an API
+# call) — this file only controls assignment, not creation.
+# ---------------------------------------------------------------------------
+
+# ── RTOS cores ──────────────────────────────────────────────────────────────
+rt:
+  - changed-files:
+    - any-glob-to-any-file:
+      - 'os/rt/**'
+      - 'test/rt/**'
+      - 'testrt/**'
+
+nil:
+  - changed-files:
+    - any-glob-to-any-file:
+      - 'os/nil/**'
+      - 'test/nil/**'
+
+oslib:
+  - changed-files:
+    - any-glob-to-any-file:
+      - 'os/oslib/**'
+      - 'test/oslib/**'
+
+# ── HAL ─────────────────────────────────────────────────────────────────────
+hal:
+  - changed-files:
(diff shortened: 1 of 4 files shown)


chibios-sheriff (commented):

**Scope:** CI/tooling — adds a mechanical-checks workflow (style + POSIX + ARM build), a path-based labeler (+ config), and a PR template that mirrors `REVIEW.md`. Targets `main`; `.github/` only, no OS/HAL code.

**Blocking:** none.
**Should-fix:** none.

**Notes (verified, non-blocking):**
- **Style gate works.** `stylecheck.py` exits 0 even on findings, but the workflow greps its output for `^(style|error):` and fails CI on a match — I verified it catches real violations. Good handling of the exit-0 quirk.
- **Fork-PR safety is correct.** `review-mechanical-checks` uses `pull_request` + `permissions: contents: read` + compile-only (no binary execution); the labeler uses `pull_request_target` but checks out the **base** config (not PR head) → no RCE. That's the safe pattern. 👍
- **Build targets verified to exist** (`test/rt/testbuild/Makefile_x86`, `test/rt-ports/testbuild/make/ARMv7-M.make`).
- Minor: labeler `sync-labels: true` removes labels that stop matching — intended, just be aware. Worth confirming every key in `labeler.yml` has a matching `gh label create`.

**Mechanical gates:** workflow YAML reviewed; not executed this pass (it runs itself on merge).

**Verdict:** advisory-LGTM — strong, well-scoped; it effectively implements the mechanical-checks gate that complements `REVIEW.md`. Human maintainer decides the merge.

gdisirio:
Thanks for this, @Ellerbach — merged. 🎉 Heads-up on two small follow-up commits I pushed to `main` after the merge, so you're not surprised when you pull:

1. **Collapsed the STM32 sub-family labels to a single `port/STM32`** (`fd17386d`). The per-family set was a bit inconsistent to maintain (U0/U3/U5, L4/L5, WB/WL were already lumped) and would need a new label + glob for every new silicon family — port-family granularity is enough for triage here. Non-port labels are unchanged.
2. **Fixed a shell-quoting nit in the labeler** (`18efc0f7`): the `oslib` line had `--description "OSLIB shared modules"--force` (no space), so `--force` got folded into the description and wasn't passed as a flag. It worked on a PR's first event but failed on every later `synchronize` with *"label already exists; use --force"*. One-space fix.

I verified both workflows end-to-end on a throwaway PR: labeling works, and the mechanical-checks style gate correctly fails on a violation while builds pass. I've also made `Review Mechanical Checks` a **required check on `main`** now. Really nice contribution — the CI gate is exactly what we wanted.

PR #6, opened 2026-06-03
Add HAL UART LLD for RP2040

Ellerbach:
## Summary

Add HAL UART LLD for RP2040

## Related

- Issue(s): no related issue but needed to have same way of using UART for .NET nanoFramework with the RP than the STM devices
- Notes for reviewers:
This has been tested (and still currently being tested in nanoFramework)
This PR to get early feedback if anything needs to be adjusted/changed so that I can move the tests in parallel

## Target Branch Check

- [x] This PR targets `main` for features/improvements
- [ ] This PR targets `stable-*` only for bugfix/regression

## Scope

## Mechanical Checks

## Testing

Currently running other tests.

## Compatibility and Risk


os/hal/ports/RP/LLD/UARTv1/hal_uart_lld.c
```diff
@@ -0,0 +1,439 @@
...
+      uartp->uart->UARTIMSC &= ~UART_UARTIMSC_TXIM;
+
+      /* Buffer transmit complete callback.*/
+      _uart_tx1_isr_code(uartp);
+
+      /* Physical end of transmission — wait for BUSY to clear would
+         require polling; signal tx2 immediately as the FIFO has been
+         fully loaded. For true physical-end detection a separate
+         mechanism would be needed.*/
+      _uart_tx2_isr_code(uartp);
```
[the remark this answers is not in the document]
Ellerbach: It's not a good implementation neither as it can block the system. And should be done differently indeed but not as a blocking code

chibios-sheriff (commented):

**Scope:** new HAL UART (callback driver model) LLD for the RP2040 port, alongside the existing SIO LLD in `UARTv1`, plus a `testhal` validation app. +1724/−2, targets `main` (feature → main ✅). Author requested early feedback — treating this as a deep pass.

**Blocking (1 — needs hardware verification):**
- **Short transmissions likely never complete — PL011 TX-interrupt "transition through level" quirk.** Per the ARM PL011 TRM (DDI 0183, UARTTXINTR), the TX interrupt asserts only when the FIFO fill level *crosses down through* the trigger level — not merely by being below it. `uart_lld_start_send()` pre-fills the FIFO and then unmasks `TXIM`; with the default `TXIFLSEL` = 1/2 (16), any send that never pushes the level *above* 16 (i.e. **n ≤ 16 from an empty FIFO**) produces no crossing → no interrupt → `_uart_tx1/tx2_isr_code` never run → `txstate` stuck in `UART_TX_ACTIVE` and `uartSendTimeout()` always returns `MSG_TIMEOUT`. The validation app masks this: its message is ~27 bytes (> 16). **Please verify on hardware with a short send (e.g. 4 bytes).** If confirmed, the "all data fitted in the FIFO" case needs explicit handling — e.g. pend the UART vector via NVIC `ISPR` after the pre-fill and let the ISR complete the TX when `txsize != 0 && txidx >= txsize` irrespective of `TXMIS`.

**Should-fix:**
- **`readme.txt` mangles existing entries:** the diff deletes two existing `NEW:` lines and replaces them with one glued line — looks like an accidental edit — and never adds an entry for the new UART driver. Restore the two lines and add a proper `- NEW: Added UART driver to the RP2040 port.` entry.
- **`txend2` is signalled immediately after `txend1`** (FIFO loaded ≠ physically transmitted; the code comment acknowledges it). Consequences: `uartSendFullTimeout()` returns early, and txend2-driven patterns (e.g. RS-485 driver-enable turn-off) would corrupt the frame tail. PL011 has no TC interrupt so this is a fair compromise — but it must be documented prominently in `hal_uart_lld.h` (driver notes), not only as a `.c` comment.
- **`uart_start()` reprograms `LCR_H` with the UART possibly still enabled** (re-`uartStart()` on a running driver). The PL011 TRM requires the UART disabled before `LCR_H` writes. Fix: clear `UARTCR` at the top of `uart_start()` before touching dividers/`LCR_H`.

**Notes (non-blocking):**
- Vector architecture verified correct: `rp_isr.c` instantiates the shared UART0/1 handlers via the `RP_UARTx_IS_USED` guards, the `HAL_USE_UART` dispatch branch already exists in `rp_uart0/1.inc`, and same-unit conflicts (SIO vs UART) are compile-time errors. Clean integration, consistent with the SIO LLD (`nvicEnableVector` usage matches port convention).
- Baud math matches the pico-sdk reference (`div = 8·clk/baud`, `IBRD = div>>7`, `FBRD = ((div&0x7F)+1)/2`) with proper range asserts and `IBRD/FBRD` latched by the `LCR_H` write. Edge nit: `FBRD` computes to 64 (6-bit field) when `(div & 0x7F) == 0x7F` — same quirk exists in pico-sdk; harmless rounding loss.
- ISR structure is sound: MIS→ICR, errors mapped to `UART_*_ERROR` flags, TX drain, RT timeout, RX drain through the official `_uart_*_isr_code` macros; `uart_enter_rx_idle_loop()` provided for the HLD completion macro. RX tail handling via RT timeout works since PL011 RT re-asserts while the FIFO is non-empty.
- `driver.mk` smart-build extension correct; headers/doxygen/license and `FALSE`-default conventions all in order.
- Suggest extending the validation app: a **short-send case** (would catch the blocking item above) and an RX path exercise (rxend/rxchar callbacks are currently all NULL).

**Mechanical gates:** CI green (style + POSIX + ARM smoke). Additionally compiled the new `testhal/RP/RP2040/UART-VALIDATION` app locally with the canonical gcc-arm-none-eabi 14.3.1 — **build OK** (CI does not build RP testhal apps, so this was the first actual compile of the LLD).

**Verdict:** **request changes (advisory)** — solid, well-integrated driver; main items are the short-send TX quirk (verify on HW) and the `readme.txt` repair. Human maintainer decides the merge.

readme.txt
```diff
@@ -94,8 +94,7 @@ See .devcontainer/README.md for included tools and usage.
 - NEW: RTC driver in XHAL.
 - NEW: Added EFL driver for simulator and a simulator MFS test application.
 - NEW: Added simulator support for x84-64 on Posix/Linux.
-- NEW: Added XSHELL_EXIT_HOOK to xshell.
-- NEW: Memory areas functions in OSLIB addressed for portability.
+- NEW: Added XSHELL_EXIT_HOOK to xshell.- NEW: Memory areas functions in OSLIB addressed for portability.
```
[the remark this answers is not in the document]
emolitor: Yes this should be fixed. Good Copilot.

readme.txt
```diff
@@ -94,7 +94,8 @@ See .devcontainer/README.md for included tools and usage.
 - NEW: RTC driver in XHAL.
 - NEW: Added EFL driver for simulator and a simulator MFS test application.
 - NEW: Added simulator support for x84-64 on Posix/Linux.
-- NEW: Added XSHELL_EXIT_HOOK to xshell.- NEW: Memory areas functions in OSLIB addressed for portability.
```
emolitor: This is a useless test comment.

emolitor (dismissed):

chibios-sheriff (commented):
🤖 **chibios-sheriff — re-review (advisory)**

**Scope:** delta `fa14249e..fd5b3527` — txend2/uartSendFullTimeout made explicitly unsupported via assertions, `readme.txt` repair. Re-review only; prior findings referenced.

**Blocking:**
- **(new) The `early` assertion breaks the asynchronous API.** `uartp->early` is only meaningful for the synchronous wrappers: `uartObjectInit()` initializes it to `false`, `uartSendTimeout()` sets `true`, `uartSendFullTimeout()` sets `false` — but the plain callback-style **`uartStartSend()` never touches it**. So `osalDbgAssert(uartp->early == true, ...)` in `uart_lld_start_send()` fires on the **first legitimate asynchronous send** in any debug build (and on any async send after a FullTimeout attempt). The LLD cannot detect "caller is uartSendFullTimeout" via `early`. Two clean exits:
  1. *(preferred)* Drop both new asserts, restore `_uart_tx2_isr_code()` right after `_uart_tx1_isr_code()`, and **document the semantics** in `hal_uart_lld.h`: on this port txend2 means "TX buffer fully loaded into FIFO", **not** line idle — so `uartSendFullTimeout()` works (returns at FIFO-load) and async works, with an explicit note that txend2 must not be used for RS-485-style bus turnaround.
  2. Keep the unsupported route, but enforce only `txend2_cb == NULL` (that assert is fine) and document that `uartSendFullTimeout()` is unsupported (it would return MSG_TIMEOUT) — the `early`-based rejection is simply not implementable from the LLD.
- **(carried, unaddressed) Short-send TX quirk** from the previous review: a send of ≤ TXIFLS trigger (16) bytes from an empty FIFO never raises `UARTTXINTR` (PL011 asserts it only on a *transition down through* the level), so the transfer never completes. The new commits do not touch the `start_send` fill/TXIM logic. Needs handling (e.g. NVIC-pend after pre-fill) or a hardware test demonstrating short sends complete.

**Should-fix:**
- `readme.txt` is better but still off: the restored lines carry stray `     |-` prefixes that do not match the file's `- NEW:` list format, and there is still **no changelog entry for the new UART driver** itself.

**Notes:** making the txend2 limitation explicit rather than silently-early is a sound direction — it just needs a mechanism that does not take the async path down with it. Test app and the rest are unchanged from the prior review.

**Mechanical gates:** CI green on `fd5b3527` (style + POSIX + ARM smoke).

**Verdict:** **request changes (advisory)** — the `early` assert is a regression that breaks `uartStartSend()` with assertions enabled; the short-send case remains open. Human maintainer decides the merge.

chibios-sheriff (commented):

Scope: re-review of `ec27cff3` + `fd5b3527` (fix round) — RP2040 UART LLD, targets `main`
Stage reached: re-verification of previously posted findings (review is staged; new deeper passes resume once these close)

**Blocking:**

1. **The short-send fix (`ec27cff3`) does not work — the pended ISR is a no-op.** The diagnosis in the new comment is exactly right (PL011 TX interrupt is transition-through-level), and pending the IRQ from `uart_lld_start_send()` is a sound mechanism. But `uart_lld_serve_interrupt()` gates the entire TX-completion path on `(mis & UART_UARTMIS_TXMIS) != 0U` — and in precisely the case being rescued, TXRIS never asserted, so the manually-pended handler reads `TXMIS == 0` and falls straight through. `_uart_tx1_isr_code()` never runs; a short `uartSendTimeout()` still hangs until timeout. Suggested shape: make TX completion state-driven rather than flag-driven, e.g.

   ```c
   if (((mis & UART_UARTMIS_TXMIS) != 0U) ||
       ((uartp->txstate == UART_TX_ACTIVE) && (uartp->txidx >= uartp->txsize))) {
   ```

   The `txstate == UART_TX_ACTIVE` guard is safe: the HLD sets `txstate` right after `uart_lld_start_send()` returns, still inside the lock zone, so the pended IRQ cannot run before it (PRIMASK held on M0+); and `_uart_tx1_isr_code()` moves `txstate` to `UART_TX_COMPLETE`, so a spurious later interrupt cannot double-complete. Easily verified on hardware with a 1–4 byte `uartSendTimeout()` from FIFO-empty.

2. **`osalDbgAssert(uartp->early == true, ...)` is still present (carried from previous review, unaddressed).** `uartObjectInit()` sets `early = false` and plain async `uartStartSend()` never touches it — so the first async send on a freshly initialized driver trips this assert in debug builds even though async sends are fully supportable by this LLD (buffer-end via `tx1` is all they need). The `txend2_cb == NULL` assert above it is the right gate to keep; this one rejects legitimate use. If the intent is to catch `uartSendFullTimeout()`, note the `early` flag cannot distinguish it from async-after-init — better to drop the assert and rely on the doc note (a `uartSendFullTimeout()` caller gets a clean MSG_TIMEOUT rather than a hang, since `tx1` won't wake a `!early` waiter).

**Should-fix:**

3. **readme.txt: the `|-` prefixes are back** (`     |- NEW: Added XSHELL_EXIT_HOOK...`). The file's convention is `- NEW:` at column 0, one entry per item — the previous concatenation should split into two plain `- NEW:` lines. Also still missing: a changelog entry for this PR's own addition (e.g. `- NEW: UART driver for RP2040 (UARTv1).`).

**Mechanical gates:** all CI green (style/POSIX/ARM smoke); `testhal/RP/RP2040/UART-VALIDATION` builds clean locally at `ec27cff3`. Note the readme issue is invisible to CI (style gate covers `.c/.h` only).

**Verdict: request changes** — items 1 and 2 are both one-line-scale fixes; happy to re-verify quickly on the next push.

chibios-sheriff (commented):

Scope: re-verification of `5f283280` ("improving robustness") — RP2040 UART LLD, targets `main`
Stage reached: re-verification of open findings

**Resolved:**

1. ✅ **Short-send completion (`txprefilled`) — verified correct.** Traced the paths: the flag is set inside the lock zone so the pended vector cannot run before the HLD finishes (`txstate` goes ACTIVE first); clearing it at the top of the ISR TX block makes completion exactly-once even when an RX interrupt on the shared vector arrives before the pended one (the RX-triggered invocation completes TX, the still-pended invocation finds the flag clear and `TXMIS`=0 and is a harmless no-op); `uart_lld_stop_send()` and `uart_lld_start()` clear it on the abort/restart paths. Nice solution — the dedicated flag is cleaner than gating on HLD state. A 1–4 byte `uartSendTimeout()` from FIFO-empty on hardware would close this definitively, if you can run one.

**Still open (carried):**

2. ❌ **`osalDbgAssert(uartp->early == true, ...)` (now line 259) — unchanged.** First plain async `uartStartSend()` on a freshly initialized driver still trips it in debug builds (`uartObjectInit()` sets `early = false`; the async API never touches it). The `txend2_cb == NULL` assert above it is the right gate to keep; this one rejects a legitimate use case. Suggest simply removing it — a `uartSendFullTimeout()` caller then gets a clean MSG_TIMEOUT instead of a wakeup, which the doc note already explains.

3. ❌ **readme.txt — unchanged:** the `     |- NEW:` prefixes at lines 97–98 (file convention is `- NEW:` at column 0, one line per entry), and the PR still has no changelog entry of its own (e.g. `- NEW: UART driver for RP2040 (UARTv1).`).

**Mechanical gates:** all CI green; `UART-VALIDATION` builds clean locally at `5f283280` (including the new `chHeapStatus()` instrumentation).

**Verdict: request changes** — item 2 is a one-line deletion, item 3 a three-line readme edit; both trivially re-verifiable.

os/hal/ports/RP/LLD/UARTv1/hal_uart_lld.c
```diff
@@ -0,0 +1,479 @@
...
+
+  /* The RP PL011 UART has no transmission-complete interrupt; physical end
+     of transmission cannot be detected without polling BUSY. txend2_cb and
+     uartSendFullTimeout() are therefore unsupported on this LLD.*/
+  osalDbgAssert(uartp->config->txend2_cb == NULL,
+                "txend2_cb not supported on RP UARTv1");
+#if UART_USE_WAIT == TRUE
+  osalDbgAssert(uartp->early == true,
+                "uartSendFullTimeout not supported on RP UARTv1");
+#endif
```
emolitor: This `early == true` assert also fires on the asynchronous `uartStartSend()` path, which never sets `early`, so a legitimate async transmit aborts in debug builds. Drop it here and reject only the `txend2_cb` case — the `uartSendFullTimeout()` case is better caught at completion time (see the comment in `uart_lld_serve_interrupt`).

```suggestion
  /* The RP PL011 UART has no transmission-complete interrupt; the physical
     end of transmission cannot be observed without polling BUSY. The "late"
     TX completion path is therefore unsupported: txend2_cb is rejected here,
     and uartSendFullTimeout() is rejected at completion time (see
     uart_lld_serve_interrupt()). The "early" path -- txend1_cb,
     uartSendTimeout() and the asynchronous uartStartSend() -- is supported.*/
  osalDbgAssert(uartp->config->txend2_cb == NULL,
                "txend2_cb not supported on RP UARTv1");
```

os/hal/ports/RP/LLD/UARTv1/hal_uart_lld.c
```diff
@@ -0,0 +1,479 @@
...
+
+      /* Buffer transmit complete callback.*/
+      _uart_tx1_isr_code(uartp);
+
+      /* Physical end of transmission (txend2_cb / uartSendFullTimeout) is
+         not supported on this LLD because the RP PL011 UART provides no
+         transmission-complete interrupt and polling BUSY from an ISR is
+         prohibited.  Callers must not configure txend2_cb or use
+         uartSendFullTimeout(); both are rejected with an assertion in
+         uart_lld_start_send().*/
```
emolitor: Catch the unsupported `uartSendFullTimeout()` here, where it can be told apart from async: at completion a waiting thread means `threadtx != NULL`, while async has none. This keeps `uartStartSend()` working while still flagging the genuinely unsupported case.

```suggestion
#if UART_USE_WAIT == TRUE
      /* uartSendFullTimeout() (early == false) blocks until the physical end
         of transmission, which this LLD cannot signal, so its waiter would
         never be resumed. The asynchronous uartStartSend() has no waiter
         (threadtx == NULL) and is fully supported; this guard therefore fires
         only on the genuinely unsupported uartSendFullTimeout() case.*/
      osalDbgAssert((uartp->early == true) || (uartp->threadtx == NULL),
                    "uartSendFullTimeout not supported on RP UARTv1");
#endif

      /* Buffer transmit complete callback.*/
      _uart_tx1_isr_code(uartp);
```

os/hal/ports/RP/LLD/UARTv1/hal_uart_lld.h
```diff
@@ -0,0 +1,303 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/**
+ * @file    UARTv1/hal_uart_lld.h
+ * @brief   RP UART subsystem low level driver header.
```
emolitor: Document the TX completion limitation on the driver itself: the PL011 has no transmission-complete interrupt, so the late path (`txend2_cb` / `uartSendFullTimeout()`) is unsupported.

```suggestion
 * @brief   RP UART subsystem low level driver header.
 *
 * @note    The RP PL011 UART has no transmission-complete interrupt, so the
 *          physical end of transmission cannot be observed. The "late" TX
 *          completion path is unsupported: @p txend2_cb is rejected and
 *          @p uartSendFullTimeout() does not complete (it times out, or
 *          blocks if called with @p TIME_INFINITE). Use the "early" path
 *          instead: @p txend1_cb, @p uartSendTimeout(), or the asynchronous
 *          @p uartStartSend().
```

readme.txt
```diff
@@ -94,8 +94,8 @@ See .devcontainer/README.md for included tools and usage.
 - NEW: RTC driver in XHAL.
 - NEW: Added EFL driver for simulator and a simulator MFS test application.
 - NEW: Added simulator support for x84-64 on Posix/Linux.
-- NEW: Added XSHELL_EXIT_HOOK to xshell.
-- NEW: Memory areas functions in OSLIB addressed for portability.
+     |- NEW: Added XSHELL_EXIT_HOOK to xshell.
+     |- NEW: Memory areas functions in OSLIB addressed for portability.
```
emolitor: These two entries got indented by mistake; restore them and add the changelog entry for this driver.

```suggestion
- NEW: Added UART driver (LLD) for the RP2040 port.
- NEW: Added XSHELL_EXIT_HOOK to xshell.
- NEW: Memory areas functions in OSLIB addressed for portability.
```

testhal/RP/RP2040/UART-VALIDATION/main.c
```diff
@@ -0,0 +1,86 @@
...
+#include "ch.h"
+#include "hal.h"
+#include "chprintf.h"
+
+#define LED_PIN              25U
+#define UART_TX_PIN          0U
+#define UART_RX_PIN          1U
+
+static const UARTConfig uartcfg = {
+  .txend1_cb  = NULL,
```
emolitor: Add an asynchronous transmit path to the test so `uartStartSend()` is actually exercised — it's the path that aborts in debug builds today. Signal completion from a `txend1_cb` callback into a binary semaphore.

```suggestion

/* Signals completion of an asynchronous (uartStartSend) transmission. */
static binary_semaphore_t txdone_sem;

static void txend1_cb(UARTDriver *uartp) {

  (void)uartp;
  chSysLockFromISR();
  chBSemSignalI(&txdone_sem);
  chSysUnlockFromISR();
}

static const UARTConfig uartcfg = {
  .txend1_cb  = txend1_cb,
```

testhal/RP/RP2040/UART-VALIDATION/main.c
```diff
@@ -0,0 +1,86 @@
...
+  msg_t status;
+  size_t heap_fragments;
+  size_t heap_free;
+  size_t heap_largest;
+  size_t n;
+  unsigned counter;
+  char txbuf[128];
+
+  halInit();
+  chSysInit();
```
emolitor, line 64: Initialize the completion semaphore (taken).

```suggestion
  chSysInit();

  chBSemObjectInit(&txdone_sem, true);
```

testhal/RP/RP2040/UART-VALIDATION/main.c
```diff
@@ -0,0 +1,86 @@
...
+  counter = 0U;
+  while (true) {
+    heap_fragments = chHeapStatus(NULL, &heap_free, &heap_largest);
+
+    n = (size_t)chsnprintf(txbuf, sizeof(txbuf),
+                           "UARTD0 validation pass %u heap_free=%u heap_largest=%u heap_fragments=%u\r\n",
+                           counter++, (unsigned)heap_free,
+                           (unsigned)heap_largest,
+                           (unsigned)heap_fragments);
+    status = uartSendTimeout(&UARTD0, &n, txbuf, TIME_MS2I(200));
```
emolitor: Alternate the synchronous and asynchronous transmit paths so both are covered. The async iteration runs with the driver's `threadtx == NULL` and a stale `early`, which is the case the LLD must accept.

```suggestion
    /* Exercise both completion paths: the synchronous uartSendTimeout() and
       the asynchronous uartStartSend() + txend1_cb. The async path runs with
       the driver's threadtx == NULL and a stale 'early' flag -- the exact
       case the LLD must accept. */
    if ((counter & 1U) == 0U) {
      status = uartSendTimeout(&UARTD0, &n, txbuf, TIME_MS2I(200));
    }
    else {
      chBSemReset(&txdone_sem, true);
      uartStartSend(&UARTD0, n, txbuf);
      status = chBSemWaitTimeout(&txdone_sem, TIME_MS2I(200));
    }
```

chibios-sheriff:
🤖 chibios-sheriff — note on @emolitor's suggestions

Verified the six suggestions — they cover all open findings, and the ISR-side guard `(early == true) || (threadtx == NULL)` is correct for all four call patterns (sync, async-after-init, async-after-sync with stale `early`, and FullTimeout which it rightly catches).

⚠️ One gap when applying them: none of the suggestions removes the **existing** `osalDbgAssert(uartp->early == true, ...)` at line 259 in `uart_lld_start_send()`. If the suggestions are applied as-is, an async `uartStartSend()` still trips that assert on entry — before emolitor's new ISR-side guard can ever do its job (the new test's async iteration would abort immediately in a debug build). Please **delete the line-259 assert** (keep the `txend2_cb == NULL` one above it) in the same push.

chibios-sheriff (commented):

Scope: re-verification of `ce0eecb3` + `8968a75c` (suggestions applied + HW testing) — RP2040 UART LLD, targets `main`
Stage reached: re-verification of open findings

**First, the CI failure is NOT this PR's fault** — the style job's changed-file detection uses a stale base sha and is picking up files changed on `main` since this PR was opened (`STM32G4xx_TEST/clocktree.h`, untouched by this PR, carries pre-existing findings). All actual PR files pass stylecheck locally. A workflow fix is being prepared on `main`; no action needed from you on this.

**Resolved:**
1. ✅ Old `early == true` assert in `uart_lld_start_send()` — deleted, async `uartStartSend()` no longer aborts. Async test coverage (txend1_cb + semaphore, alternating paths) applied — good.
2. ✅ readme: the driver's own changelog entry is present (line 97).
3. ✅ `txprefilled` path reported hardware-hammered (`ce0eecb3`) — closes the HW-verify recommendation.

**New blocking:**
4. ❌ **`_uart_tx2_isr_code()` is now fired at buffer-end** (`uart_lld_serve_interrupt`, after `_uart_tx1_isr_code`). tx2 means *physical end of transmission*; at that point the FIFO has merely been *loaded* — up to 17 frames (16 FIFO + shifter) are still on the wire, ~1.8 ms at 115200. Consequences: `txend2_cb` fires early and `uartSendFullTimeout()` returns while the line is still transmitting — silently wrong for exactly its main use cases (RS-485 direction turnaround, half-duplex). This also contradicts the PR's own `hal_uart_lld.h` note, which still says txend2 "is rejected" — the deleted `txend2_cb == NULL` assert was the right gate and should come back. Honest-unsupported beats silently-early: please remove the `_uart_tx2_isr_code()` call, restore the `txend2_cb == NULL` assert, and (optionally but recommended) add @emolitor's ISR-side guard `(early == true) || (threadtx == NULL)` to catch `uartSendFullTimeout()` cleanly.

**Should-fix:**
5. Line endings: `8968a75c` converted `hal_uart_lld.c` wholesale to CRLF (and `hal_uart_lld.h` was created CRLF). Code files in this tree are LF (`os/rt/src/chsys.c`, `os/hal/ports/RP/RP2040/hal_lld.h`, ...) — please normalize both LLD files to LF; the CRLF churn also destroys line-history for the whole file.
6. readme line 100: one `     |- NEW:` prefix still remains (`Memory areas functions...`) — its sibling was fixed, this one was missed.

**Mechanical gates:** UART-VALIDATION builds clean locally at `8968a75c`; POSIX/ARM CI green; style CI red for the unrelated reason above.

**Verdict: request changes** — item 4 is the only substantive one (delete one call, restore one assert); 5 and 6 are mechanical.

chibios-sheriff:
🤖 chibios-sheriff — offer

@Ellerbach the three remaining items are small and fully specified (remove the `_uart_tx2_isr_code()` call + restore the `txend2_cb == NULL` assert, normalize the two LLD files to LF, fix the last readme `|-` line). If you prefer, I can push those fixes directly to this branch as sheriff commits — you'd just pull before continuing. Say the word and I'll do it; otherwise they're all yours. Either way the final review/approval stays with the human maintainers.

emolitor (approved):

Ellerbach:
@gdisirio there was still one more fix to go :-) Improvements from your sheriff!

PR #9, opened 2026-06-06
Improvements for RP2040 HAL UART LLD

Ellerbach:
## Summary

Improvements for RP2040 HAL UART LLD

## Related

- Issue(s): few more things from #6 after sheriff feedback
- Notes for reviewers:

## Target Branch Check

      merged on `main`

## Scope

## Mechanical Checks

## Testing

## Compatibility and Risk


os/hal/ports/RP/LLD/UARTv1/hal_uart_lld.c
```diff
@@ -251,10 +251,14 @@ void uart_lld_stop(UARTDriver *uartp) {
...
 
   /* The RP PL011 UART has no transmission-complete interrupt; physical end
-     of transmission cannot be detected without polling BUSY. The txend2_cb
-     callback is therefore unsupported on this LLD.*/
+     of transmission cannot be detected without polling BUSY. txend2_cb and
+     uartSendFullTimeout() are therefore unsupported on this LLD.*/
   osalDbgAssert(uartp->config->txend2_cb == NULL,
                 "txend2_cb not supported on RP UARTv1");
+#if UART_USE_WAIT == TRUE
+  osalDbgAssert((uartp->early == true) || (uartp->threadtx == NULL),
```
chibios-sheriff: This guard cannot fire here. In the HLD, `uartSendFullTimeout()` sets `early = false`, calls `uart_lld_start_send()`, and only **after** it returns suspends the caller via `osalThreadSuspendTimeoutS(&uartp->threadtx, ...)` — so `threadtx` is still NULL at this point for *every* API, and the condition is always true.

With the completion-time copy removed, the unsupported `uartSendFullTimeout()` is no longer detected at all: its waiter is never resumed (`_uart_wakeup_tx1_isr()` only resumes when `early == true`), so the call silently hangs until its timeout — forever with `TIME_INFINITE`. (UART-VALIDATION exercises `uartSendTimeout()`/`uartStartSend()` only, so a local test run cannot show this.)

At submission time the LLD genuinely cannot tell `uartSendFullTimeout()` (`early == false`, `threadtx == NULL`) from an asynchronous `uartStartSend()` issued after init (identical state). The two only become distinguishable at completion, when the FullTimeout waiter is suspended (`threadtx != NULL`) — that is why the assert sat in `uart_lld_serve_interrupt()`. Please restore it there; this copy can stay or go (dead code, but harmless).

readme.txt
```diff
@@ -98,6 +98,7 @@ See .devcontainer/README.md for included tools and usage.
 - NEW: Added UART driver (LLD) for the RP2040 port.
 - NEW: Added XSHELL_EXIT_HOOK to xshell.
 - NEW: Memory areas functions in OSLIB addressed for portability.
+
```
chibios-sheriff: Stray blank line — the `*** Next ***` list has no separator lines; please drop it.

chibios-sheriff (dismissed):

Scope: RP2040 UART LLD follow-up — relocates the `uartSendFullTimeout()` guard from the ISR completion path into `uart_lld_start_send()`, comment touch-ups, readme. Target: `main`.
Stage reached: C (deep)

> not the full review yet.

Blocking:
- The relocated `(early == true) || (threadtx == NULL)` assert can never fire at submission time, and with the completion-time copy removed the unsupported `uartSendFullTimeout()` case is no longer detected anywhere — its waiter silently hangs until timeout. Details in the inline comment; please restore the guard in `uart_lld_serve_interrupt()`.

Notes:
- readme.txt: stray blank line inserted in the middle of the `*** Next ***` changelog list (inline comment).
- The rewritten comment block in `uart_lld_serve_interrupt()` indents continuation lines with 8 spaces instead of aligning under the comment text like the surrounding code — cosmetic, can ride along with the fix.

Mechanical gates: builds green (RP2040 UART-VALIDATION + POSIX simulator); `stylecheck.py` clean on the changed `.c`.

Verdict: request changes (restore the completion-time guard).

chibios-sheriff:
The completion-time guard relocation is exactly right — that resolves the blocker. I noticed the `tfl changes` commit had moved the `tools/ftl` submodule back to an older ancestor (`92a392ac`), unrelated to the UART work, so I pushed a small commit (`a7370ca0`) resetting it to main's current pointer (`6357e6e9`). Looks like an accidental stale submodule pointer — nothing for you to do, just flagging it. Thanks!

gdisirio (approved):
(posted by the assistant on the maintainer's instruction)

PR #13, opened 2026-06-07
OTGv1: keep the OTG1 ULPI clock gate disabled on STM32H7

chibios-sheriff:
🤖 chibios-sheriff — PR authored from a forum bug report (advisory; review and merge decisions stay with human maintainers)

## Summary

Fixes the WFI/sleep hang with an active OTG1 (USB) on STM32H7: the OTG1 instance (RM0433 "USB2OTG") has its own ULPI clock gates — `USB2OTGHSULPIEN` / `USB2OTGHSULPILPEN`, the latter **enabled after reset** — which the driver never touched, so the unclocked-but-enabled ULPI interface prevented entering/leaving sleep mode. This is the same problem historically worked around for OTG2 (forum t=1798), on the instance that workaround does not cover.

A new H7-only `rccDisableOTG_FSULPI` alias is cleared in the OTG1 start path; on platforms without an OTG1 ULPI gate the code compiles away (guarded by `#if defined(...)`).

Note: the ChibiOS OTG1/OTG2 indices are reversed with respect to RM0433's USB1/USB2 naming — the commit message documents the mapping.

Origin: forum bug report by **geebee** — https://forum.chibios.org/viewtopic.php?t=6696. Register facts verified against RM0433 (RCC_AHB1LPENR bit 28 reset state; the RM's revision history shows these bits were added to the manual after the original port was written).

## Related

- Issue(s): forum bug report https://forum.chibios.org/viewtopic.php?t=6696
- Notes for reviewers:
  - Compile-verified and `#warning`-probe-verified (path present on H743, absent on F407); the sleep-mode behavior itself still wants a hardware confirmation on a dual-OTG H7 board — the reporter could test this branch.
  - Out of scope, recorded: full `BOARD_OTG1_USES_ULPI` support (HS/ULPI PHY on OTG1 for the H7 models whose second instance has the ULPI option) — feature work, maintainer's call.

## Target Branch Check

      merged on `main`

## Scope

- [x] I reviewed impact on other ports/configurations where relevant (alias defined only in the STM32H7XX block; F4/F7/F2/L4/F105 paths unchanged, verified by building F407/F746)

## Mechanical Checks

- [x] Style check passed on changed `.c`/`.h` files (changed lines clean; 2 pre-existing findings untouched)
- [x] Build check passed (ARM target compiles, see Testing)

## Testing

- [x] Test run successfully locally: `testhal/STM32/multi/USB_CDC` for H743 (OTG1+OTG2), H723, F407, F746
- Any other tests run on a device? Not yet — hardware confirmation of the sleep behavior pending (see reviewer notes)

## Compatibility and Risk

- [x] Risks/limitations are documented below: behavior change only on H7+OTG1, where the ULPI gate is now explicitly kept off; any (unsupported) configuration relying on the reset-enabled gate would need the new alias removed — none exists in-tree


gdisirio:
We could use a HW test for confirmation on this one.

gdisirio (approved):

PR #17, opened 2026-06-09
SYSTICKv1: add STM32_ST_FREQUENCY_TOLERANCE for the free running tick

chibios-sheriff:
🤖 Prepared by the ChibiOS sheriff (assistant-operated account) on the maintainer's direction.

Stage 1 of the fix for the STM32U0/U3 system-tick clock-rounding issue found while verifying #16.

### Problem
In free running mode, `st_lld_init()` required the ST clock to be an **exact** integer multiple of `OSAL_ST_FREQUENCY` (`osalDbgAssert((ST_CLOCK_SRC % OSAL_ST_FREQUENCY) == 0)`). STM32U0/U3 expose **accurate, non-round** MSI frequencies when the MSI feeds the PLL (the "4 MHz" range is **3.998 MHz**, via the `msipllrange[]` table), so an MSI-sourced SYSCLK like **55.972 MHz** is not a multiple of a 10 kHz tick → the assertion halts the system at boot when `CH_DBG_ENABLE_ASSERTS` is on.

### Change
Mirroring the established `STM32_USB_48MHZ_DELTA` pattern, add **`STM32_ST_FREQUENCY_DELTA`** — the allowed ST tick deviation in **per-mille**. The prescaler is now rounded to the nearest integer and the achieved tick is checked against the tolerance.

- **Default is 0** → exact divisor still required → **no behavior change** for existing configs.
- A device/app sets e.g. `5` (0.5%) to allow a small rounding.
- SYSTICKv1 only (SYSTICKv2 is RTC-based — separate path).

### Verification (hardware, NUCLEO-U083RC)
- Default `0` + asserts on → halts in `st_lld_init` (unchanged).
- `STM32_ST_FREQUENCY_DELTA = 5` + asserts on → boots, RTC alarm A/B callbacks fire normally.
- Exact-clock SYSTICKv1 target (G474) builds/runs unaffected; stylecheck clean on changed lines.

### Follow-up (separate)
Stage 2/3: set `STM32_ST_FREQUENCY_DELTA` (0.5%) in the **U0/U3 mcuconf templates** (chibios-ftl) and regenerate the affected in-tree configs.

### DIFF
--- a/os/hal/ports/STM32/LLD/SYSTICKv1/hal_st_lld.c
+++ b/os/hal/ports/STM32/LLD/SYSTICKv1/hal_st_lld.c
@@ -318,15 +318,18 @@
 #error "STM32_ST_USE_TIMER specifies an unsupported timer"
 #endif
 
-#if 0 /* TODO remove */
-#if ST_CLOCK_SRC % OSAL_ST_FREQUENCY != 0
-#error "the selected ST frequency is not obtainable because integer rounding"
-#endif
-
-#if (ST_CLOCK_SRC / OSAL_ST_FREQUENCY) - 1 > 0xFFFF
-#error "the selected ST frequency is not obtainable because TIM timer prescaler limits"
-#endif
-#endif
+/* Free running prescaler rounded to the nearest integer, the resulting tick
+   frequency and its absolute deviation from OSAL_ST_FREQUENCY. These are
+   expressions only (no objects): they fold to constants on static-clock
+   devices and, where used solely inside osalDbgAssert(), they fully vanish
+   when assertions are disabled.*/
+#define ST_PSC          ((((halfreq_t)ST_CLOCK_SRC +                         \
+                           ((halfreq_t)OSAL_ST_FREQUENCY / 2U)) /            \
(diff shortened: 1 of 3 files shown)


chibios-sheriff:
Updated: the rounding/tolerance math is now expressed as pure macros (`ST_PSC`/`ST_TICK`/`ST_TICK_ERROR`) used only inside the `osalDbgAssert()` calls, so the check leaves **no generated code when assertions are disabled** — only the prescaler value reaches the register write, and it folds to a compile-time constant on static-clock devices. Verified: U083RC builds with no warnings (asserts off, default delta 0) and still boots+runs with asserts on + `STM32_ST_FREQUENCY_DELTA=5`; G474 (exact clock) unaffected.

gdisirio (approved):

gdisirio (approved):

chibios-sheriff:
Renamed the setting to `STM32_ST_FREQUENCY_TOLERANCE` (a percentage-style tolerance reads better than "delta", which suits a fixed-frequency check like USB). Unit unchanged: per-mille, default 0.

PR #12, opened 2026-06-07
STM32C0xx: add SPI2 RCC macros and DMAMUX identifiers

chibios-sheriff:
🤖 chibios-sheriff — PR authored from a forum bug report (advisory; review and merge decisions stay with human maintainers)

## Summary

Adds the missing SPI2 support pieces in the STM32C0xx port: `rccEnableSPI2()`/`rccDisableSPI2()`/`rccResetSPI2()` in `stm32_rcc.h` and the SPI2 DMAMUX request identifiers (18/19) in `stm32_dmamux.h`. The registry already enables SPI2 on C051/C071/C091/C092, but `STM32_SPI_USE_SPI2` failed to compile with exactly these identifiers missing.

Origin: forum bug report by **geoffrey.brown** (https://forum.chibios.org/viewtopic.php?t=6679), who also supplied the RCC macros.

Verification: RCC bits checked against the CMSIS header (`RCC_APBENR1_SPI2EN`, `RCC_APBRSTR1_SPI2RST`); DMAMUX inputs 18/19 checked against RM0490 Table 49 and the STM32G0xx port. A C071 application with `STM32_SPI_USE_SPI2=TRUE` builds clean on this branch and fails on unpatched `main` with the three missing identifiers.

## Related

- Issue(s): forum bug report https://forum.chibios.org/viewtopic.php?t=6679
- Notes for reviewers:
  - The other half of the report is in the FTL templates: `mcuconf_stm32c071xx/mcuconf.h.ftl` lists only SPI1 — follow-up PR planned in the chibios-ftl repository.
  - Recorded, out of scope: `stm32_dmamux.h` is also missing I2C2 (12/13) and USART3/4 (54–57) identifiers used by newer C0 family members (C071 registry already claims `STM32_HAS_I2C2 TRUE`) — candidate for a separate audit/fix.

## Target Branch Check

- [x] This PR targets `main` — all changes land on `main` first (upstream-first policy);
      `stable-*` branches only receive maintainer-selected backports of commits already
      merged on `main`

## Scope

- [x] Change is focused and does not bundle unrelated work
- [x] I reviewed impact on other ports/configurations where relevant (STM32C0xx only; additive macros)

## Mechanical Checks

- [x] Style check passed on changed `.c`/`.h` files
- [x] Build check passed (ARM target compile, see Testing)

## Testing

- [x] Test run successfully locally: C071 build with SPI driver + SPI2 enabled (positive), unpatched-tree negative proof
[...]

Issue #3, opened 2026-06-02
chThdCreateFromMemoryPool() rejects valid fixed memory pools due to overly strict alignment assertion


gdisirio:
Description
  chThdCreateFromMemoryPool() currently asserts that mp->align >=
  PORT_WORKING_AREA_ALIGN before allocating an object from the pool.

  This is too strict for fixed/preloaded memory pools initialized with
  chPoolObjectInit() and later loaded with externally allocated or statically
  allocated objects that are already aligned to PORT_WORKING_AREA_ALIGN.

  In that case, mp->align records the pool’s declared/default alignment, often
  PORT_NATURAL_ALIGN, but it does not necessarily describe the actual alignment
  of the preloaded objects. The actual stack safety condition is already checked
  immediately after allocation:

  chDbgAssert(MEM_IS_ALIGNED(wbase, PORT_WORKING_AREA_ALIGN),
              "invalid pool object alignment");

  Example
  On ARMv6-M:

  #define PORT_NATURAL_ALIGN       sizeof(void *)  /* 4 */
  #define PORT_WORKING_AREA_ALIGN  8U

  A fixed pool can be initialized with natural alignment:

  chPoolObjectInit(&pool, THD_WORKING_AREA_SIZE(stack_size), NULL);

  and then loaded with correctly aligned storage:

  storage = chHeapAllocAligned(NULL, pool_size, PORT_WORKING_AREA_ALIGN);
  chPoolLoadArray(&pool, storage, object_count);

  Calling chThdCreateFromMemoryPool(&pool, ...) then halts at:

  chDbgAssert(mp->align >= PORT_WORKING_AREA_ALIGN, "invalid pool alignment");

  even though the returned object would satisfy MEM_IS_ALIGNED(wbase,
  PORT_WORKING_AREA_ALIGN).

  Impact
  Valid fixed/preloaded pools cannot be used for dynamic thread creation on
  ports where PORT_WORKING_AREA_ALIGN > PORT_NATURAL_ALIGN, such as ARMv6-M.
  This causes debug/assert builds to halt despite the actual working area being
  correctly aligned.

  Proposed Fix
  Only require mp->align >= PORT_WORKING_AREA_ALIGN for provider-backed/growing
  pools, where mp->align is passed to the provider and therefore defines the
  alignment contract for future allocations.

  chDbgAssert((mp->provider == NULL) ||
              (mp->align >= PORT_WORKING_AREA_ALIGN),
              "invalid pool alignment");

[...]

PR #19, opened 2026-06-09
configs: tree-wide mcuconf/xmcuconf regeneration pass

chibios-sheriff:
Regenerates **every templated in-tree `mcuconf.h`/`xmcuconf.h`** (298 configs) via `tools/updater` to bring them back in sync with their generation templates, and bumps `tools/ftl` to the F303 I2S template fix (chibios-ftl #3, merged).

**Verified loss-free** — a full per-file audit found **zero removed `#define`s**. The changes:

| Change | Configs |
|---|---|
| `STM32_ST_FREQUENCY_TOLERANCE` added (default `0`, no behaviour change) | 298 |
| STM32F3 I2S driver section restored (template-gap fix propagating) | 18 |
| stale comment fix `xhalconf.h` → `halconf.h` | 4 |

**Loss handling:** the only family where regen would have dropped settings was F303 — its template was missing the I2S section that the SPIv2 LLD (`hal_i2s_lld.c/.h`) and the F303 configs use. Fixed the template (chibios-ftl #3) and re-regenerated, so the I2S settings are preserved. Every other config is pure add-only.

**Excluded from the pass:**
- the experimental `*_clocktree` configs (`u385`/`g474` nucleo64) — different `STM32_CFG_*` clock-tree scheme, not this template;
- the two `stm32c071rb_nucleo64` demo configs — owned by the held PR #12 (not pre-empted here; #12 should be rebased on this once unheld so it picks up the tolerance line).

**Verification:** stylecheck clean on all 298; F3 / C071 / U3 targets build clean. Regen is idempotent against the merged templates.

Sheriff-authored; for human review/merge.

### DIFF
--- a/demos/STM32/CMSIS-STM32F407-DISCOVERY/cfg/mcuconf.h
+++ b/demos/STM32/CMSIS-STM32F407-DISCOVERY/cfg/mcuconf.h
@@ -298,6 +298,7 @@
  */
 #define STM32_ST_IRQ_PRIORITY               8
 #define STM32_ST_USE_TIMER                  2
+#define STM32_ST_FREQUENCY_TOLERANCE        0
 
 /*
  * UART driver system settings.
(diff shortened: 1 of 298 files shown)


gdisirio (approved):
Changes look good, regeneration is loss-free. (posted by the assistant on the maintainer's instruction)

PR #33, opened 2026-06-12
Port context setup refactor and MPU switched regions as shared table pointers

gdisirio:
First batch from the SB improvements track (see `os/sb/note_sb_improvement_priorities.md`, added by this PR; design rationale in `os/sb/note_svc_mpu_optimizations.md` point 3 on `main`).

## What this changes

**1. `PORT_SETUP_CONTEXT()` becomes the `port_setup_context()` inline function (all ports)**

The port-dependent part of thread creation is now an inline function taking a pointer to the port-owned `struct port_context` instead of the thread pointer. The `&tp->ctx` step happens at the call sites (RT `chthreads.c`, NIL `ch.c`) where the thread types are complete — this is what dodges both reasons the macro existed (header ordering, RT/NIL thread-type polymorphism). The macro, its per-port helper sub-macros, and the `chport.h` existence check are removed. Pure refactor, behavior identical; groundwork for the next commit, where context setup needs real conditional code.

**2. MPU switched regions become a shared table pointer (ALT ports)**

The thread context no longer embeds the switched MPU region values; it holds a pointer to a const regions table:

- Sandbox threads point at a per-SB table in `sb_class_t` (filled once at start); all other threads share a const default table in flash with all switched regions disabled.
- Switch-out store-back is removed entirely — the tables are the master copy, the MPU is never read back.
- Switch-in compares old/new table pointers and skips the MPU when equal: kernel↔kernel and SB-sibling switches never touch the MPU; reprogramming happens only on protection-domain crossings.
- The load is a single ldm/stm burst: ARMv7-M through the RBAR/RASR alias registers with VALID|region# pre-baked in the stored RBAR values (no RNR writes); ARMv8-M as RNR plus up to four RBAR/RLAR pairs per burst.
- Memory: 8 bytes/region/thread replaced by one pointer per thread plus one table per protection domain.

**3. `port_setup_context_base()` becomes a mandatory port function (all ports)**

Initializes the context fields that must be valid for thread objects representing already-running execution flows (the boot thread of each instance, which never goes through the full creation path). Needed because removing the store-back turned the regions pointer into the one read-before-write context field. Empty in ports that don't need it; `port_setup_context()` invokes it rather than duplicating the initialization; RT/NIL call it unconditionally.

## Verification

Built clean: G474 SB_HOST_SWITCHED + SB_HOST_NOSWITCH (ARMv7-M-ALT), H563 SB_HOST_SWITCHED (ARMv8-M-ML-ALT), L552 + L552 TZ_HOST (ARMv8-M-ML, -TZ), RT/NIL ARMCM4, ARMCM0, ARMCR8, ARM7 generics, POSIX simulator (both 32/64-bit, SIMIA32/SIMX86_64), SB guest apps (msh). Disassembly of both SB demos checked instruction-by-instruction against the intended switch sequences.

Not built (no toolchain or no in-tree consumer): AVR, e200, ARMvx-M-SB, templates — converted by inspection.

### Hardware (2026-06-16 — G474RE Nucleo, ARMv7-M-ALT)

`RT-STM32G474RE-NUCLEO64-SB_HOST_SWITCHED` (`PORT_SWITCHED_REGIONS_NUMBER = 2`) flashed and run with both guests — SB1 = `SB-CLIENT-RT-XHAL-XSHELL` @ `0x08040000`, SB2 = `SB-CLIENT-BARE-METAL` @ `0x08060000`:

- **SB1**: interactive XShell over the VCP; ran the full in-sandbox ChibiOS/RT + oslib test suite → **`Final result: SUCCESS`**, zero failures. This drives the host↔sandbox switch path continuously.
- **SB2**: confirmed live via GDB (`state = CH_STATE_WTMSG`, not `FINAL`; no fault/restart loop — CPU idle when halted, system quiescent).
- **MPU design confirmed on-chip** (GDB): each sandbox's `ctx.regions` points at its own `sb_class_t.mpuregs` table, the two SB tables are distinct and both differ from the flash-resident `port_mpu_default_regions`; stored RBARs carry the pre-baked `VALID | region#` with bases matching each guest's flash/RAM layout; the switched regions read back disabled while in kernel/idle (default table loaded).

**Still pending:** H563 SB_HOST_SWITCHED (ARMv8-M-ML-ALT) on hardware — build/disassembly-verified only, no board attached.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gdisirio (commented):
_(Posted under the wrong account by mistake. The first-layer review is reposted below under **@chibios-sheriff** — please read that one.)_

chibios-sheriff (commented):

**Scope:** `PORT_SETUP_CONTEXT()` → `port_setup_context()` refactor (all ports) + ALT-port MPU switched-regions as a shared table pointer + mandatory `port_setup_context_base()`. Base `main`.
**Stage reached:** D (deep review C completed; style D clean)

> not the full review yet.

**Stage A — build (local, GCC 14.3.1):** clean on every touched Cortex-M family + sim:
- ARMv7-M-ALT: G474 SB_HOST_SWITCHED + SB_HOST_NOSWITCH
- ARMv8-M-ML-ALT: H563 SB_HOST_SWITCHED
- ARMv7-M: F407-DISCOVERY · ARMv6-M: L053R8 · SIM: RT-Posix-Simulator

(ARMv7-R / e200 / AVR / classic-ARM use other toolchains — not built here.)

**Stage C — deep review:** I checked the new switch path against `struct port_context` and it holds up. The asm offsets are correct in all FPU/`PORT_SAVE_CONTROL` combinations (v7 `CONTEXT_MPU_OFFSET` = `offsetof(regions)` = 44/48/108/112; v8 `MPU_OFFSET` unchanged from before); `r5` (callee-saved) safely carries the outgoing context base from store to `PORT_LOAD_MPU_CONTEXT` across the scheduler block; the skip-when-equal table-pointer compare is sound; the SB per-domain fill matches each port's burst (v7 pre-bakes `VALID|REGION(i)`, v8 leaves region selection to the RNR writes); and `port_setup_context_base()` + the default table correctly cover the boot thread's read-before-write `regions` field. **No correctness defect found.**

**Should-fix:**
- **Weld the asm offsets to the struct.** `CONTEXT_INTEGER_SIZE` and `CONTEXT_MPU_OFFSET`/`MPU_OFFSET` in `chcoreasm.S` are hand-coded constants that must track `struct port_context`. They're correct today, but nothing enforces it — a later field add or an FPU/`CONTROL` layout change would silently desync them and the switch would load the MPU from a wrong offset (a bogus `regions` pointer → a memory-protection failure that compiles clean and CI can't catch). A `_Static_assert(offsetof(struct port_context, regions) == …)` (plus one for the integer-block size) in each ALT `chcore.c` would harden exactly the seam this PR introduces.
- **Changelog.** No top-level `readme.txt` (`*** Next ***`) entry — the switched-regions change is user-visible (lower per-thread footprint; MPU reprogrammed only on protection-domain crossings). The `os/common/ports/readme.txt` API note is good but separate.

**Notes:**
- Build-clean can't exercise the asm offsets or the skip-logic; the gating evidence is an on-hardware run with real thread switches incl. an SB protection-domain crossing (kernel↔SB and SB-sibling). The description cites clean builds — worth confirming a HW run before merge.
- Scope: the PR also carries four SB design-note `.md` files + several notes commits — reasonable as the sandboxes-dev track's groundwork, just flagging.

**Mechanical gates:** `stylecheck.py` clean on changed lines; builds above pass.

**Verdict: comment** — no blocking items. Happy to draft the `_Static_assert` guards and/or the changelog line if useful.

gdisirio:
@chibios-sheriff We need to stay with C99, static assertions are beyond that.

chibios-sheriff:
🤖 chibios-sheriff

Fair correction — `_Static_assert`/`static_assert` is C11, so on the C99 baseline that suggestion doesn't apply; please disregard it.

If a compile-time guard is ever wanted within C99, the negative-array idiom (`typedef char x[(cond) ? 1 : -1];`) is an option — but it only welds things cleanly if the offset constant lives in a shared header visible to both `chcore.c` and `chcoreasm.S` (today `CONTEXT_MPU_OFFSET`/`MPU_OFFSET` are `#define`s inside the `.S`). That's a small refactor, not a one-liner, so it's entirely your call — I'm not pushing it; the finding was non-blocking. For the record the offsets check out as-is (I verified `CONTEXT_MPU_OFFSET` == `offsetof(struct port_context, regions)` across all FPU/CONTROL combinations). The only remaining should-fix is the top-level `readme.txt` `*** Next ***` changelog entry.

PR #43, opened 2026-06-17
SB: guard the VRQ-state atomicity invariant (re-review of margin 6)

gdisirio:
## Summary

Re-review and close Tier‑1 item 2 (VRQ masking atomicity) from the SB improvement priorities.

The fastcall handlers in `host/sbvrq.c` perform unlocked read‑modify‑write on the shared VRQ state (`vrq.isr`, `vrq.wtmask`, `vrq.flags[]`). This confirms those sequences are atomic **by priority design, not by luck**, and adds a build‑time guard so the invariant can't silently regress.

## Why it's safe

- Handlers run at **SVCall** priority.
- The only VRQ producers (`sbVRQTriggerI`, `sbVRQSetFlagsI`) are I‑class, so they can only run from an IRQ at `CORTEX_MAX_KERNEL_PRIORITY` or below — which the ALT ports pin **one level under SVCall** (`CORTEX_MAX_KERNEL_PRIORITY == CORTEX_PRIORITY_SVCALL + 1`).
- A producer therefore cannot preempt a handler mid‑sequence, closing:
  - the `wtmask` lost‑update (dropped VRQ),
  - the `flags[]` read‑and‑clear race in `gcsts` vs `sbVRQSetFlagsI` (the VIO completion path),
  - the `isr = 0` → mask‑check window in `vrq_enable`/`vrq_return`.
- The fast‑priority band above SVCall can't produce VRQs (no OS primitives); `sbVRQTriggerS` runs under `chSysLock`, same‑core only.

No locking is needed.

## Change

- `host/sbvrq.c`: a `#if … #error` guard asserting `CORTEX_PRIORITY_SVCALL < CORTEX_MAX_KERNEL_PRIORITY` (C99‑safe preprocessor; inert on ports lacking the symbols), with the rationale in a comment.
- Documentation: isolation note margin 6 updated to "safe by design + guarded", priorities item 2 and the open‑points host bullet marked resolved.

Verified: builds clean on the ARMv8‑M‑ML‑ALT host demo; guard proven active (inverting the condition fires the `#error`).

### DIFF
--- a/os/sb/host/sbvrq.c
+++ b/os/sb/host/sbvrq.c
@@ -32,6 +32,28 @@
 /* Module local definitions.                                                 */
 /*===========================================================================*/
 
+/* The read-modify-write sequences on the shared VRQ state (vrq.isr,
+   vrq.wtmask, vrq.flags[]) in the fastcall handlers below are NOT explicitly
+   locked: their atomicity against the asynchronous producers (sbVRQTriggerI()
+   and sbVRQSetFlagsI(), both I-class) rests entirely on a priority invariant.
+   The handlers run at the SVCall priority, while any IRQ allowed to call an
+   I-class function sits at CORTEX_MAX_KERNEL_PRIORITY or below (numerically
+   higher, i.e. less urgent). SVCall therefore out-prioritizes every VRQ
+   producer and a producer cannot preempt a handler mid-sequence; the
+   sequences are atomic by construction (see note_sb_isolation_security.md,
+   margin 6). The fast-priority band above SVCall cannot produce VRQs because
+   it cannot call OS primitives.
+
+   This guard fails the build if a port or priority reconfiguration ever
+   breaks that ordering, rather than silently reopening the lost-update race.
+   The ALT ports define CORTEX_MAX_KERNEL_PRIORITY == CORTEX_PRIORITY_SVCALL + 1
+   so the invariant holds. (Lower numeric value == higher priority.) */
+#if defined(CORTEX_PRIORITY_SVCALL) && defined(CORTEX_MAX_KERNEL_PRIORITY)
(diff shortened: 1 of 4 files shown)


chibios-sheriff (commented):

**Scope:** build-time guard for the VRQ-state atomicity invariant in `os/sb/host/sbvrq.c` (+ doc updates). Base `main`.
**Stage reached:** D (full A→D pass, clean)

> not the full review yet.

**Stage A — build (GCC 14.3.1):** clean — `RT-STM32H563ZI-…-SB_HOST_SWITCHED` (ARMv8-M-ML-ALT) and `RT-STM32G474RE-…-SB_HOST_SWITCHED` (ARMv7-M-ALT).

**Stage C — deep:** the argument holds. I verified the invariant against the actual port headers: both `ARMv7-M-ALT` and `ARMv8-M-ML-ALT` define `CORTEX_MAX_KERNEL_PRIORITY = CORTEX_PRIORITY_SVCALL + 1` — SVCall is strictly more urgent (lower number) than every I-class VRQ producer, so a producer (`sbVRQTriggerI`/`sbVRQSetFlagsI`) cannot preempt a handler mid read-modify-write on `vrq.isr`/`vrq.wtmask`/`vrq.flags[]`. The unlocked sequences are atomic by construction, and the `#if … #error` correctly fails the build if `SVCALL >= MAX_KERNEL_PRIORITY` ever holds. C99-safe (pure preprocessor) and inert where the symbols aren't defined (`#if defined(...)` wrapper) — good portability hygiene.

**Stage D — style:** clean on changed lines.

**Notes:**
- No `readme.txt` entry needed — this refines the SB VRQ machinery still under `*** Next ***` (unreleased) and adds no functional change.

**Mechanical gates:** SB host builds pass (v7-M-ALT + v8-M-ALT); stylecheck clean on changed lines.

**Verdict: advisory-LGTM** — correctness verified against the port headers; no findings. (Human approval/merge as always.)

PR #51, opened 2026-06-23
Fix registry thread reference accounting

gdisirio:
Summary:
- Count thread references consistently in registry lookup/addref paths even when dynamic threads are disabled.
- Keep OSAL thread release behavior aligned with the updated reference accounting.

Testing:
- Covered by the RT testbuild assertion and coverage runs performed during the RT coverage work.

Back-port note:
- This fixes a reference-accounting footgun that also exists on 21.11.x; it should be back-ported to the 21.11.x maintenance branch as a separate PR.

### DIFF
--- a/os/common/abstractions/nasa_cfe/osal/src/osapi.c
+++ b/os/common/abstractions/nasa_cfe/osal/src/osapi.c
@@ -1903,7 +1903,8 @@ int32 OS_TaskDelete(uint32 task_id) {
   funcptr_t fp;
 
   /* Check for thread validity, getting a reference.*/
-  if (chRegFindThreadByPointer(tp) == NULL) {
+  tp = chRegFindThreadByPointer(tp);
+  if (tp == NULL) {
     return OS_ERR_INVALID_ID;
   }
 
@@ -1947,7 +1948,8 @@ int32 OS_TaskWait(uint32 task_id) {
   thread_t *tp = (thread_t *)task_id;
 
   /* Check for thread validity, getting a reference.*/
-  if (chRegFindThreadByPointer(tp) == NULL) {
+  tp = chRegFindThreadByPointer(tp);
+  if (tp == NULL) {
     return OS_ERR_INVALID_ID;
   }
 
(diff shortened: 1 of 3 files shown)


chibios-sheriff (commented):

**Scope:** RT thread-registry reference-accounting fix (+ an OSAL `OS_TaskGetInfo` deref-before-validate fix). Base `main`.
**Stage reached:** D (full A→D pass, clean)

> not the full review yet.

**Stage A — build:** clean — POSIX-sim RT both with the default config **and with `CH_CFG_USE_DYNAMIC=FALSE`** (the exact path the fix targets) links cleanly.

**Stage C — deep:** verified.
- `thread_t.refs` is always present (chobjects.h) and `chThdRelease()` is always compiled (only its dispose path is `CH_CFG_USE_DYNAMIC`-guarded), so the unconditional `refs++`/`chThdRelease` is sound on non-dynamic builds.
- The real bug: with dynamic disabled, `chRegFirstThread()`/`chRegNextThread()` skipped the addref while `chThdRelease()` still decremented → reference-count underflow (and the `chDbgAssert(refs > 0)` would trip). Making the registry lookups reference unconditionally aligns them with the always-on release. Correct.
- OSAL ref balance holds: all four touched `OS_Task*` functions release the reference they take — `SetPriority`/`GetInfo` via `chThdRelease`, `Delete`/`Wait` via `chThdWait`. Using the returned referenced pointer (`tp = chRegFindThreadByPointer(tp)`) is equivalent-or-safer.
- `OS_TaskGetInfo()`: moving the `wasize = …tp->wabase…` computation *after* the validity check fixes a genuine deref-before-validate on a bogus task id. Nice.

**Stage D — style:** clean on changed lines.

**Changelog:** I added the `*** Next ***` entries to the branch (`073c32c6`) — this fixes *released* RT/OSAL behaviour, so per the reviewer-owned changelog rule it warrants them (registry refcount + the `OS_TaskGetInfo` deref).

**Notes:**
- You flagged a **21.11.x backport** — I can prep that as a separate PR once this merges (same flow as the NVIC fix).

**Mechanical gates:** sim builds (default + dynamic=FALSE); stylecheck clean on changed lines.

**Verdict: advisory-LGTM** — refcount alignment and OSAL balance verified, builds clean on the targeted non-dynamic config. (Human merge as always.)

PR #58, opened 2026-06-29
MFS: fix verification and mount checking edge cases

gdisirio:
## Summary
- fix MFS write verification when the source aliases the shared non-cacheable buffer
- honor MFS_CFG_STRONG_CHECKING during mount scans
- prevent duplicate transaction erases from corrupting used-space accounting

## Testing
- testhal/simulator/EFL-MFS default build and run
- testhal/simulator/EFL-MFS with MFS_CFG_STRONG_CHECKING=FALSE

## Backport
A backport to stable-21.11.x is needed.

### DIFF
--- a/os/hal/lib/complex/mfs/hal_mfs.c
+++ b/os/hal/lib/complex/mfs/hal_mfs.c
@@ -542,9 +542,9 @@ static mfs_error_t mfs_bank_scan_records(MFSDriver *mfsp,
      header.*/
   while (hdr_offset < end_offset - ALIGNED_DHDR_SIZE) {
     mfs_data_header_t dhdr;
-    uint16_t crc;
     flash_offset_t data_offset;
     flash_offset_t data_available;
+    bool valid;
 
     /* Reading the current record header.*/
     RET_ON_ERROR(mfs_flash_read(mfsp, hdr_offset,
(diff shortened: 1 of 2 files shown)


chibios-sheriff (commented):

Scope: MFS verification / mount-scan / transaction edge-case fixes (`hal_mfs.c`, `hal_mfs.h`). Target: `main`.

> not the full review yet.

All four changes verified:
- **Write-verify aliasing** — when `wp` aliases `ncbuf`, the read-back now targets the upper buffer half (`vbuf = &data8[BUFFER_SIZE/2]`, `vmax = BUFFER_SIZE/2`) selected by an integer (`uintptr_t`) range test — correctly sidestepping the UB of relational comparison between pointers into different objects — plus a lower-half-fit `osalDbgAssert` and the new `BUFFER_SIZE >= 32` `#error`. Fixes the buffer-against-itself comparison that always passed.
- **`mfs_flash_copy`** `copy_unit` halved under `WRITE_VERIFY` so staged data stays in the lower half, disjoint from the read-back scratch — consistent with the above.
- **`MFS_CFG_STRONG_CHECKING`** now gates the mount-time CRC read-back (was unconditional). Default is `TRUE`, so default builds are unchanged; `FALSE` accepts on metadata and defers data-integrity to read time.
- **`mfsCommitTransaction`** erase subtracts `used_space` only when `descriptors[i].offset != 0`, so a duplicate in-transaction erase of one id no longer corrupts the accounting.

**Mechanical gates:** Stage A — `testhal/simulator/EFL-MFS` builds **and runs `Final result: SUCCESS`** both default and with `MFS_CFG_STRONG_CHECKING=FALSE` (latter set via `mfsconf.h`). Note: setting it via a command-line `UDEFS=` override trips an unrelated SIM-link error in `ch_test.c` on strict linkers — project-side build-infra, not this PR. Stage D style clean on changed lines.

**Changelog (reviewer-owned):** no `readme.txt` entry yet; drafted for `*** Next ***`:
```
- FIX: MFS write verification could pass spuriously when the source data
       aliased the shared non-cacheable buffer; the read-back now uses a
       disjoint buffer half. MFS_CFG_STRONG_CHECKING is now honored during
       mount scans, and a duplicate in-transaction erase no longer corrupts
       the used-space accounting (github PR #58).
```
Say the word and I'll push it; it'll gain `(backported to 21.11.6)` once the stable backport you flagged lands.

Verdict: **advisory-LGTM** (advisory — a human merges), pending the `readme.txt` entry.

PR #59, opened 2026-06-29
ROMFS: guard chunked reads and seeks

gdisirio:
Summary:
- Guard chunked ROMFS reads at EOF and reject zero-progress/over-reporting chunk backends.
- Validate SEEK_CUR and SEEK_END bounds before signed vfs_offset_t addition.

Validation:
- git diff --check
- make -C demos/STM32/RT-STM32H563ZI-NUCLEO144-SB_HOST_SWITCHED -j4

Note:
- The raw lwIP web-server demo could not be built in this checkout because ext/lwip/src/Filelists.mk is missing.

### DIFF
--- a/os/vfs/drivers/romfs/drvromfs_impl.inc
+++ b/os/vfs/drivers/romfs/drvromfs_impl.inc
@@ -243,6 +243,10 @@ static ssize_t romfs_chunked_read(void *session, vfs_offset_t offset,
     return CH_RET_EINVAL;
   }
 
+  if ((offset == descp->size) || (n == 0U)) {
+    return 0;
+  }
+
   transferred = 0U;
   while (n > 0U) {
     size_t chunk_index;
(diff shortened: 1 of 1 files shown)


chibios-sheriff (commented):

Scope: ROMFS chunked-read EOF/backend guards + `SEEK_CUR`/`SEEK_END` bounds (`drvromfs_impl.inc`). Target: `main`.

> not the full review yet.

Verified:
- `romfs_chunked_read`: early `return 0` on EOF (`offset == descp->size`) or `n == 0`; and a misbehaving chunk backend is now rejected with `EINVAL` — `copied == 0` (would spin forever) and `copied > chunk_available` (would overrun `buf`). Both are real defensive wins.
- `__romfile_setpos_impl` `SEEK_CUR`/`SEEK_END`: bounds checked before the signed addition. Confirmed `vfs_offset_t`, `self->size` and `self->position` are all `int32_t` (signed), so `(vfs_offset_t)0 - self->position` and `self->size - self->position` are well-defined signed arithmetic — no unsigned-wrap trap. Inclusive bounds are correct (`SEEK_END` `0`→`size`, `-size`→`0`; `SEEK_CUR` confined to `[0, size]`).

**Mechanical gates:** Stage A — `RT-STM32H563ZI-NUCLEO144-SB_HOST_SWITCHED` (a ROMFS/VFS user) builds clean (GCC 14.3.1). Stage D style clean.

**Changelog:** not needed — this refines the VFS/ROMFS integration still under `*** Next ***` (unreleased, covered by the existing `NEW: SBs and VFS integration` line; unreleased-feature exception, REVIEW.md §5). If ROMFS actually shipped in a prior release, say so and I'll add a `FIX` entry.

PR #61, opened 2026-06-29
C++ wrappers: fix synchronization and container helpers

gdisirio:
Summary:
- Fix Scheduler::rescheduleS() so it actually calls chSchRescheduleS().
- Correct message polling context handling and registry reference handoff.
- Prevent copying wrappers around embedded kernel objects.
- Fix typed mailbox fetch/peek paths for small T values.
- Round ObjectsPool storage stride and fix ThreadsPool access to its MemoryPool.

Validation:
- git diff --check
- make -C demos/STM32/RT-STM32F407-DISCOVERY-G++ -j4
- temporary compile-only probe instantiating ThreadsQueue, semaphores, mutexes, events, Mailbox<uint8_t>, ObjectsPool<uint8_t>, and ThreadsPool

### DIFF
--- a/os/various/cpp_wrappers/ch.hpp
+++ b/os/various/cpp_wrappers/ch.hpp
@@ -407,7 +407,7 @@ namespace chibios_rt {
      */
     static void rescheduleS(void) {
 
-      void chSchRescheduleS();
+      chSchRescheduleS();
     }
   };
 
@@ -584,6 +584,8 @@ namespace chibios_rt {
    *            an object of this type.
    */
   class ThreadReference final {
+    friend class Registry;
+
     /**
      * @brief   Pointer to the system thread.
      */
@@ -756,6 +758,25 @@ namespace chibios_rt {
      */
     bool isPendingMessage(void) const {
 
+      bool pending;
+
+      chSysLock();
+      pending = chMsgIsPendingI(thread_ref);
+      chSysUnlock();
+
+      return pending;
+    }
+
+    /**
+     * @brief   Returns true if there is at least one message in queue.
+     *
+     * @retval true             A message is waiting in queue.
+     * @retval false            A message is not waiting in queue.
+     *
+     * @iclass
+     */
+    bool isPendingMessageI(void) const {
+
       return chMsgIsPendingI(thread_ref);
     }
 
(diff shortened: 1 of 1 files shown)


chibios-sheriff (commented):

Scope: C++ wrappers correctness fixes (`os/various/cpp_wrappers/ch.hpp`). Target: `main`.

> not the full review yet.

Each fix verified:
- **`Scheduler::rescheduleS()`** — `void chSchRescheduleS();` was a function *declaration* (a no-op) → now an actual call. Real bug.
- **`ThreadReference` messages** — `isPendingMessage()` now wraps the I-class `chMsgIsPendingI()` in `chSysLock/Unlock`; added an `isPendingMessageI()` for I-context callers.
- **`Registry::nextThread()`** — now takes `ThreadReference&` and nulls the caller's inner pointer. `chRegNextThread()` *consumes* (releases) the passed thread's reference, so the wrapper must not retain and re-release it; this fixes a stale-reference handoff. (Minor API change by-value→by-ref, but necessary and correct; uses a new `friend class Registry`.)
- **Copy/assignment deleted** on `SynchronizationObject` (base of `ThreadsQueue`), `EventListener`, `EventSource`, `MailboxBase` — embedded kernel objects must not be value-copied. The base gains an explicit default ctor (required once a copy op is user-declared); `ThreadsQueue`'s own deletes are correctly dropped (now inherited).
- **Typed `Mailbox<T>` for small T** — the important one: the old `reinterpret_cast<msg_t*>(msgp)` made `chMBFetch*` write a full `msg_t` through a `T*`, **corrupting memory when `sizeof(T) < sizeof(msg_t)`**. Now it fetches into a local `msg_t` and narrows only on `MSG_OK`; `post*`/`peek` use value casts.
- **`ObjectsPool`** — `object_size` rounded up to `sizeof(void *)` (each free slot stores a next-pointer), fixing stride/alignment for small `T`; `ThreadsPool` granted `friend` access to `MemoryPool`.

**Mechanical gates:** Stage A — `RT-STM32F407-DISCOVERY-G++` builds clean (g++ 14.3.1). Stage D — `ch.hpp` is a `.hpp`, which the CI stylecheck does not scan (it globs `*.c`/`*.h`); the few hits `stylecheck.py` reports on changed lines are false positives (it reads C++ `operator=` / `= delete` as a "glued assignment operator"). No real style issue.

**Changelog (reviewer-owned):** drafted a `FIX` entry for `*** Next ***`. Say the word and I'll push it — or, if `os/various/cpp_wrappers` isn't changelog-tracked, I'll drop it.
```
- FIX: C++ wrappers (ch.hpp): Scheduler::rescheduleS() now actually
       reschedules; typed Mailbox<T> fetch/peek no longer overrun the
       caller's storage when T is smaller than msg_t; ObjectsPool rounds its
       object size up to pointer alignment; the synchronization wrappers are
       no longer copyable and the registry reference handoff is corrected
       (github PR #61).
```

Verdict: **advisory-LGTM** (advisory — a human merges), pending the changelog decision.

PR #62, opened 2026-06-29
RT: remove unused thread descriptor dispose parameter

gdisirio:
## Summary
- remove the unused tdispose parameter from deprecated RT thread descriptor macros
- update the internal static-thread descriptor call and the USELIB header mirror

## Testing
- rg -n "tdispose" os demos test
- make -C test/rt/testbuild -j$(nproc)
- ./build/ch from test/rt/testbuild

Note: please check stable-21.11.x too to see whether this cleanup should be applied there.


chibios-sheriff (commented):

Scope: RT — drop the unused `tdispose` parameter from the `@deprecated` thread-descriptor macros (`chthreads.h`/`.c` + the `RT-ARMCM4-USELIB` `ch.h` mirror). Target: `main`.

> not the full review yet.

Verified:
- `tdispose` was genuinely unused — `__THD_DESC_DATA`'s body never referenced it (no `.dispose` field, just a trailing comma after `.owner`), so removing it from `__THD_DESC_DATA`/`THD_DESC_DECL` and the trailing `NULL` from `THD_DESCRIPTOR`/`_AFFINITY` is mechanical and behavior-neutral. The one internal caller (`chThdCreateStatic`) and the USELIB `ch.h` mirror are updated consistently. `THD_DESCRIPTOR`'s public 6-arg form is unchanged; only the deprecated `THD_DESC_DECL`/`__THD_DESC_DATA` shed the no-op arg.

**Mechanical gates:** Stage A — `test/rt/testbuild` builds and runs `Final result: SUCCESS` (rt + oslib suites); `RT-ARMCM4-USELIB` demo builds. Stage D style clean.

On your stable question: `stable-21.11.x` has no `tdispose`/`__THD_DESC_DATA` at all (these macros are kernel-8/main-only) → **no 21.11.x backport applicable**.

**Changelog:** not needed — the thread "dispose" feature these macros belong to is itself unreleased (`*** Next ***`: "NEW: capability to associate a dispose function to threads"), so this is pre-release refinement (unreleased-feature exception).

PR #63, opened 2026-06-29
RT: make working-area reuse checks atomic

gdisirio:
## Summary
- add an I-class boolean registry scan for working-area presence without touching references
- move thread working-area reuse assertions into locked creation paths, immediately before registry insertion
- remove the non-atomic public-wrapper checks that used the reference-taking registry finder

## Testing
- git diff --check
- make -C test/rt/testbuild -j$(nproc)
- ./build/ch from test/rt/testbuild

Note: please check stable-21.11.x too and apply the same fix there if affected.

### DIFF
--- a/os/rt/include/chregistry.h
+++ b/os/rt/include/chregistry.h
@@ -128,6 +128,7 @@ extern "C" {
   thread_t *chRegNextThread(thread_t *tp);
   thread_t *chRegFindThreadByName(const char *name);
   thread_t *chRegFindThreadByPointer(thread_t *tp);
+  bool chRegIsWorkingAreaInUseI(stkline_t *wa);
   thread_t *chRegFindThreadByWorkingArea(stkline_t *wa);
 #ifdef __cplusplus
 }
--- a/os/rt/src/chregistry.c
+++ b/os/rt/src/chregistry.c
@@ -283,6 +283,35 @@ thread_t *chRegFindThreadByWorkingArea(stkline_t *wa) {
   return NULL;
 }
 
+/**
+ * @brief   Confirms that a working area is being used by some active thread.
+ *
+ * @param[in] wa        pointer to a static working area
+ * @retval true         if a matching thread has been found.
+ * @retval false        if a matching thread has not been found.
+ *
+ * @iclass
+ */
+bool chRegIsWorkingAreaInUseI(stkline_t *wa) {
+  ch_queue_t *tqp;
+
+  chDbgCheckClassI();
+
+  /* Scanning registry.*/
+  tqp = REG_HEADER(currcore)->next;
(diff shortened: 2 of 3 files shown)


chibios-sheriff (commented):

Scope: RT — make the working-area-reuse assertion atomic (new ref-free I-class scan + checks moved into the locked creation paths). `chregistry.h`/`.c`, `chthreads.c`. Target: `main`.

> not the full review yet.

Verified — the rationale holds and the implementation is correct:
- The old check `chDbgAssert(chRegFindThreadByWorkingArea(wbase) == NULL, …)` ran in the public wrappers **before** `chSysLock` (non-atomic), and `chRegFindThreadByWorkingArea` iterates via `chRegFirstThread()`/`chRegNextThread()` — the **reference-taking** iterators — so it churns reference counts (and would leak a reference on a match). Wrong primitive, and a TOCTOU window.
- New `chRegIsWorkingAreaInUseI()`: a raw `REG_HEADER(currcore)` walk (`rqueue`/`__CH_OWNEROF`, comparing `chThdGetWorkingAreaX(ctp) == wa`), **no references touched**, `chDbgCheckClassI()` — the correct I-class boolean primitive.
- The assertion now lives inside the locked/I-class creation paths immediately before `REG_INSERT`: `chThdSpawnSuspendedI`, `chThdCreateSuspendedI`, and the locked section of `chThdCreateStatic`. I verified **every** creation entry point routes through one of these — `chThdSpawnRunningI`→`chThdSpawnSuspendedI`, `chThdCreateI`→`chThdCreateSuspendedI`, and the public wrappers call the `*I` forms under lock — so coverage is preserved on all paths and is now atomic (and applies uniformly under `CH_CFG_USE_REGISTRY`, no longer gated on `STACK_CHECK||DYNAMIC`). `chRegIsWorkingAreaInUseI` is only evaluated inside `chDbgAssert`, so zero cost when asserts are disabled.

**Mechanical gates:** Stage A — `test/rt/testbuild` builds and runs `Final result: SUCCESS` both default and with `CH_DBG_SYSTEM_STATE_CHECK`+`CHECKS`+`ASSERTS` (exercises the new I-class assert at runtime). Stage D style clean on all three files.

On your stable question: **yes, `stable-21.11.x` is affected** — it has the same pre-lock `chRegFindThreadByWorkingArea` asserts (`chthreads.c:229/289/337`) and, after the #54 backport, ref-taking iterators. A backport is warranted; it'll need adaptation (stable uses `threadref`/`stkalign_t` and has no `__CH_OWNEROF`). Flagging it as a backport candidate for your call.

**Changelog:** this refines the registry reference-accounting rework (#51), still in `*** Next ***` (unreleased), so I'd fold it there and add no separate entry — unless you want the atomicity fix called out, in which case I'll add a `FIX` line.

PR #72, opened 2026-07-03
Add Makefile rules to build .uf2 images for RP MCUs as part of the main make target ("all")

electroniceel:
## Summary

Adds Makefile rules to build .uf2 images for RP MCUs as part of the main make target ("all").
These Makefile rules are then used in the two demo projects for the RP2040 and RP2350.

## Target Branch Check

      merged on `main`

## Scope

## Mechanical Checks

## Testing

- [x] Demo .uf2 for the RP2040 uploaded and ran successfully
- [x] Demo .uf2 for the RP2350 uploaded and ran successfully

## Compatibility and Risk

Risk:

Building the  RP2040 and RP2350 demo projects now requires the `picotool` binary, as distributed by Raspberry Pi at https://github.com/raspberrypi/picotool. In the past this was not strictly required just for building. But for uploading the binary to a Pico or Pico2 the integrated bootloader is the most common way, and for this users already needed picotool. If this is really an issue, the new make rules file can just be commented out from the demo project Makefile.

So I think this change is acceptable.

### DIFF
--- a/os/hal/ports/RP/rp_uf2_image.mk
+++ b/os/hal/ports/RP/rp_uf2_image.mk
@@ -1,0 +1,24 @@
+# make rules for building an .uf2 image, suitable for use with the bootloader on the RP MCUs
+
+# these make rules should be included _BEFORE_ $(RULESPATH)/rules.mk
+
+# path to the picotool binary, can be overridden
+# picotool is distributed as part of the Raspberry Pi Pico SDK
+# https://github.com/raspberrypi/pico-sdk
+PICOTOOL ?= picotool
+
+# check if picotool binary is available & executable
+$(if $(shell $(PICOTOOL)), , $(error picotool not found. Download it from https://github.com/raspberrypi/picotool\
+ and either install it in $$PATH or set the PICOTOOL variable to the location))
+
+# build a .uf2 file out of the .elf to use with the bootloader of the RP MCUs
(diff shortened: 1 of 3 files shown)


electroniceel:
@emolitor could you please review this change?

os/hal/ports/RP/rp_uf2_image.mk
```diff
@@ -0,0 +1,24 @@
...
+
+# these make rules should be included _BEFORE_ $(RULESPATH)/rules.mk
+
+# path to the picotool binary, can be overridden
+# picotool is distributed as part of the Raspberry Pi Pico SDK
+# https://github.com/raspberrypi/pico-sdk
+PICOTOOL ?= picotool
+
+# check if picotool binary is available & executable
+$(if $(shell $(PICOTOOL)), , $(error picotool not found. Download it from https://github.com/raspberrypi/picotool\
```
emolitor, line 11: I will rework this, its a *HUGE* red flag to automatically attempt to download tools which is why a previous similar version of this was rejected.

emolitor: Ahhh cleaning up now. Reworked so if picotool is installed and on the path UF2 is created. But if picotool is not installed UF2 is not created but the build doesn't fail. We shouldn't require picotool as it is optional. However I agree that if its in the path the devex will be better to generate the UF2.

PR #84, opened 2026-07-15
stm32: restore standalone TIM14 GPT ISR

PhysicistJohn:
## Summary

Restore the generic standalone `STM32_TIM14_HANDLER` in the STM32 TIMv1 GPT low-level driver. STM32F0 platforms already provide the handler and IRQ number, but enabling `STM32_GPT_USE_TIM14` currently stops at `#error "TIM14 ISR not defined by platform"`.

The handler remains under the existing `STM32_GPT_USE_TIM14` and `!STM32_TIM14_SUPPRESS_ISR` guards and dispatches `GPTD14` using the standard OSAL IRQ prologue/epilogue.

## Related

- Issue(s): None.
- Notes for reviewers: this is independent of the USB PMA allocator finding. A stable-21.11.x backport, if desired, is maintainer-selected.

## Target Branch Check

- [x] This PR targets `master`, the current/default integration branch.

## Scope

- [x] Shared-vector and application-suppressed ISR configurations are unchanged.

## Mechanical Checks

- [x] `tools/style/stylecheck.pl` passed on the changed C file.
- [x] POSIX simulator build passed.
- [x] STM32F072RB NUCLEO64 ARM build passed with GPT and TIM14 enabled.

## Testing

- [x] Current unpatched `master` reproduces the documented compile-time error.
- [x] Patched build succeeds and its ELF contains `GPTD14`.
- [x] The same correction is integrated in reproducible tinySA F072 and F303 ChibiOS 21.11.5 builds.
- Hardware callback timing was not tested on F072 hardware.

## Compatibility and Risk

- [x] No intentional public API/ABI break.
- The change only supplies the previously missing generic handler when TIM14 GPT is enabled and not suppressed.

### DIFF
--- a/os/hal/ports/STM32/LLD/TIMv1/hal_gpt_lld.c
+++ b/os/hal/ports/STM32/LLD/TIMv1/hal_gpt_lld.c
@@ -406,7 +406,17 @@ OSAL_IRQ_HANDLER(STM32_TIM8_UP_HANDLER) {
 
 #if STM32_GPT_USE_TIM14 || defined(__DOXYGEN__)
 #if !defined(STM32_TIM14_SUPPRESS_ISR)
-#error "TIM14 ISR not defined by platform"
+#if !defined(STM32_TIM14_HANDLER)
+#error "STM32_TIM14_HANDLER not defined"
+#endif
+OSAL_IRQ_HANDLER(STM32_TIM14_HANDLER) {
+
+  OSAL_IRQ_PROLOGUE();
+
+  gpt_lld_serve_interrupt(&GPTD14);
+
+  OSAL_IRQ_EPILOGUE();
+}
 #endif /* !defined(STM32_TIM14_SUPPRESS_ISR) */
 #endif /* STM32_GPT_USE_TIM14 */
 


chibios-sheriff (commented):

Scope: restore a standalone `OSAL_IRQ_HANDLER(STM32_TIM14_HANDLER)` for GPT in the TIMv1 LLD (`hal_gpt_lld.c`), targeting `master`.
Stage reached: B (structural)

> not the full review yet.

Thanks for this, and for the thorough reproducer notes. The underlying need is real: on STM32F0 (and F1/F37x) TIM14 has its own dedicated interrupt vector, yet enabling `STM32_GPT_USE_TIM14` currently stops at the `#error`, so TIM14-as-GPT is unusable there today.

The finding is structural — about *where* the handler should live.

**Should-fix (architecture).** ChibiOS has two timer-ISR mechanisms:

- The **modern, centralized** one — `os/hal/ports/STM32/<family>/stm32_isr.[ch]` pulling in the per-timer `LLD/TIMv1/stm32_timXX.inc` dispatchers. `stm32_tim14.inc` already provides a single `OSAL_IRQ_HANDLER(STM32_TIM14_HANDLER)` that fans out to whichever driver uses TIM14 (GPT / ICU / PWM / ST) and resolves the shared-vector cases (e.g. F4 routes TIM14 through the combined `TIM8_TRG_COM_TIM14` handler). Ports on this path set `STM32_TIM14_SUPPRESS_ISR`, which suppresses the per-driver handlers below.
- The **legacy per-driver** one — the `OSAL_IRQ_HANDLER` blocks in `hal_gpt_lld.c` / `hal_pwm_lld.c` / `hal_icu_lld.c`. The TIM9–14 `#error` stubs are placeholders in this path.

The reason F0/F1/F37x hit the `#error` is that **those older ports were never migrated to the centralized `stm32_isr` / `.inc` mechanism** — they still rely on the legacy per-driver handlers. This PR patches that legacy path, which has three issues:

1. It's **GPT-only and TIM14-only** — the identical gap exists for ICU and PWM on TIM14 (and for TIM9–13) on the same ports, so it addresses one corner of a port-wide gap.
2. It **bypasses the collision handling** the `.inc` dispatcher exists to provide — it compiles on F0 only because TIM14 happens to be standalone there, so the pattern doesn't generalize to the shared-vector families.
3. If F0 is later modernized (adopting `stm32_tim14.inc`, which unconditionally defines `OSAL_IRQ_HANDLER(STM32_TIM14_HANDLER)`), this handler becomes a **duplicate symbol**.

The right fix is to bring the F0 port (and eventually F1/F37x) onto the centralized `stm32_isr.[ch]` + `stm32_timXX.inc` infrastructure and set the suppress flags accordingly — the same shape C0/G0 already use.

**We'd like to keep this PR open** rather than close it: the maintainer would prefer to use it as the starting point for patching F0 support the modern way. So please read this as a direction note, not a rejection. If you're up for it, converting the F0 `stm32_isr` to the `.inc` timer dispatchers (TIM14 first) would be very welcome and we can help steer it; otherwise it's a good marker for the maintainer to pick up.

Mechanical gates: not re-run this pass — the finding is structural and settles direction before a build matrix is worth spending. Your reported F072 build is consistent with the legacy path compiling in isolation.

Verdict: comment — kept open for the F0 `stm32_isr` modernization approach; not suitable to merge in its current (legacy-path) form.

chibios-sheriff:
To unblock you **today without any tree change**, the driver already supports exactly this case through the `SUPPRESS_ISR` hook — you provide the TIM14 ISR from your application and own the vector. On F0 this works cleanly because TIM14 has a dedicated vector (`Vector8C`, IRQ 19) and isn't used as the OS tick.

Recipe (F072, GPT on TIM14):

1. In `mcuconf.h`, enable the timer and suppress the driver-side ISR:
   ```c
   #define STM32_GPT_USE_TIM14                 TRUE
   #define STM32_TIM14_SUPPRESS_ISR            /* defined → no #error, and the
                                                  driver will not touch the NVIC
                                                  vector; that's now yours */
   ```

2. Provide the handler in your application (any `.c` that includes `hal.h`):
   ```c
   OSAL_IRQ_HANDLER(STM32_TIM14_HANDLER) {

     OSAL_IRQ_PROLOGUE();

     gpt_lld_serve_interrupt(&GPTD14);

     OSAL_IRQ_EPILOGUE();
   }
   ```
   `STM32_TIM14_HANDLER` resolves to the correct vector for the part, and `gpt_lld_serve_interrupt()` is public (declared in `hal_gpt_lld.h`).

3. Enable the NVIC vector yourself once the driver is started (the `SUPPRESS_ISR` guard also disables the driver's own `nvicEnableVector`/`nvicDisableVector` for TIM14):
   ```c
   gptStart(&GPTD14, &gptcfg);
   nvicEnableVector(STM32_TIM14_NUMBER, STM32_GPT_TIM14_IRQ_PRIORITY);
   ```
   On F072 `STM32_TIM14_NUMBER` is 19 and `STM32_GPT_TIM14_IRQ_PRIORITY` defaults to 7 (override in `mcuconf.h` if you like). Pair it with `nvicDisableVector(STM32_TIM14_NUMBER)` if/when you stop the driver.

That gives you working TIM14 GPT on F0 right now, entirely in your own project, with no fork of the tree. The PR stays open as the tracking item for the proper long-term fix (bringing the F0 `stm32_isr` onto the centralized `.inc` dispatchers), which is the maintainer's call.

PR #85, opened 2026-07-15
stm32: preserve EP0 PMA across reconfiguration

PhysicistJohn:
## Summary

`usb_lld_disable_endpoints()` preserves endpoint zero but reset the PMA allocation cursor to the descriptor-table boundary. Rebuilding a configuration could then allocate endpoint 1 over the still-live EP0 buffers.

Reset the allocator and immediately reserve the configured EP0 IN and OUT sizes before initializing data endpoints. Apply the same ownership rule to both maintained STM32 PMA drivers. The full bus-reset path remains a true allocator reset before EP0 initialization.

## Related

- Issue(s): None.
- Notes for reviewers: this is independent of the STM32F0 TIM14 change in #84. Stable backports are maintainer-selected.

## Target Branch Check

- [x] This PR targets `master`, the current/default integration branch.

## Scope

- [x] USBv1 and USBv2 allocator and endpoint-disable semantics were reviewed together.
- [x] Isochronous aliases still reserve one underlying buffer, matching endpoint initialization.

## Mechanical Checks

- [x] ChibiOS style check passed independently for both changed C files.
- [x] POSIX simulator build passed.
- [x] STM32F303 USB CDC build passed through USBv1.
- [x] STM32G0B1 USB CDC build passed through USBv2.

## Testing

- The USBv1 defect is reproduced by an exact 21.11.5 firmware ELF: after a second `SET_CONFIGURATION(1)`, EP0 TX and EP1 TX both own PMA address `0x0040`.
- With this USBv1 correction, the same executable scenario passes `1 -> 1`, `1 -> 0 -> 1`, CDC traffic, suspend/wakeup, EP0 STALL, and final bus-reset re-enumeration with distinct active PMA addresses.
- The corrected USBv1 image also enumerated and transferred CDC data on STM32F303 hardware.
- USBv2 has matching source semantics and current-master build coverage; USBv2 runtime hardware was not tested.

## Compatibility and Risk

- [x] No intentional public API/ABI break.
- Full bus reset is unchanged and still initializes EP0 from a clean allocator.
- Only the endpoint-disable path that keeps EP0 active retains its packet-memory reservation.
- Disabling same-value configuration rebuilds is not used as a workaround; explicit `1 -> 0 -> 1` is covered.

### DIFF
--- a/os/hal/ports/STM32/LLD/USBv1/hal_usb_lld.c
+++ b/os/hal/ports/STM32/LLD/USBv1/hal_usb_lld.c
@@ -121,6 +121,26 @@ static uint32_t usb_pm_alloc(USBDriver *usbp, size_t size) {
 }
 
 /**
+ * @brief   Resets the packet memory allocator while preserving EP0 buffers.
+ * @details Endpoint zero remains active when the other endpoints are
+ *          disabled, therefore its packet memory cannot be made available to
+ *          a subsequently initialized endpoint.
+ *
+ * @param[in] usbp      pointer to the @p USBDriver object
+ */
+static void usb_pm_reset_after_ep0(USBDriver *usbp) {
+  const USBEndpointConfig *epcp = usbp->epc[0];
+
+  usb_pm_reset(usbp);
+  if (epcp->in_state != NULL) {
+    (void)usb_pm_alloc(usbp, epcp->in_maxsize);
+  }
+  if (epcp->out_state != NULL) {
+    (void)usb_pm_alloc(usbp, epcp->out_maxsize);
+  }
+}
+
+/**
  * @brief   Reads from a dedicated packet buffer.
  *
  * @param[in] ep        endpoint number
@@ -682,8 +702,9 @@ void usb_lld_init_endpoint(USBDriver *usbp, usbep_t ep) {
 void usb_lld_disable_endpoints(USBDriver *usbp) {
   unsigned i;
 
-  /* Resets the packet memory allocator.*/
-  usb_pm_reset(usbp);
+  /* Resets the packet memory allocator without releasing the still-active
+     endpoint-zero buffers.*/
+  usb_pm_reset_after_ep0(usbp);
 
   /* Disabling all endpoints.*/
   for (i = 1; i <= USB_ENDPOINTS_NUMBER; i++) {
--- a/os/hal/ports/STM32/LLD/USBv2/hal_usb_lld.c
+++ b/os/hal/ports/STM32/LLD/USBv2/hal_usb_lld.c
@@ -215,6 +215,26 @@ static uint32_t usb_pm_alloc(USBDriver *usbp, size_t size) {
 }
 
 /**
+ * @brief   Resets the packet memory allocator while preserving EP0 buffers.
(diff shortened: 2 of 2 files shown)


chibios-sheriff (commented):

Scope: preserve the EP0 packet-memory reservation across a configuration rebuild in the STM32 USBv1 and USBv2 LLDs (`hal_usb_lld.c` ×2), targeting `master`.

> not the full review yet.

Nice find — this is a real, subtle correctness bug and the fix is well reasoned.

**Correctness (Stage C) — confirmed.** `usb_lld_disable_endpoints()` called `usb_pm_reset()`, rewinding the PMA cursor to the BTABLE boundary while EP0 stays active, so the next `usb_lld_init_endpoint(1)` after a `SET_CONFIGURATION` could allocate EP1 over the still-live EP0 buffers. The new `usb_pm_reset_after_ep0()` resets and immediately re-reserves EP0 IN/OUT, which:
- never NULL-derefs `epc[0]`: the only caller is `usbDisableEndpointsI()`, which clears `epc[1..N]` but not `epc[0]`, and the bus-reset path (`usb_lld_reset()`) sets `epc[0] = &ep0config` with its own untouched `usb_pm_reset()` — that path is deliberately left as a full allocator reset;
- re-reserves in the **same order and sizes** (`in_maxsize` then `out_maxsize`) that `usb_lld_init_endpoint(0)` uses, so the cursor lands exactly past EP0's real footprint and EP1+ no longer overlap;
- is applied symmetrically to both maintained PMA drivers (USBv1 + USBv2).

**Structural (Stage B) — clean.** Correct layer, self-contained static helper, doxygen present, bus-reset semantics untouched.

**Mechanical gates:**
- Build (Stage A): `testhal/STM32/multi/USB_CDC` links clean on `stm32f303_discovery` (USBv1) and `stm32g0b1re_nucleo64` (USBv2), ARM GCC 14.3.1.
- Style (Stage D): `tools/style/stylecheck.py` clean on both changed files.
- The change cherry-picks cleanly onto current `master` (the two files are untouched there since this branch was cut).

**Note (non-blocking):** the branch is a fair way behind `master` (branched before a large batch of unrelated merges), so please **rebase onto current `master`** before merge — it applies without conflict in the driver files, but `readme.txt` has moved.

**Changelog (reviewer-owned).** This fixes released behaviour, so it takes a `readme.txt` entry. Because the branch predates recent `readme.txt` changes, it's cleanest to add this on rebase rather than onto the stale base — drafted here for that:

```
- FIX: STM32 USBv1 and USBv2 packet memory could be double-allocated across a
       configuration rebuild. usb_lld_disable_endpoints() reset the PMA
       allocator to the descriptor-table boundary while endpoint zero remained
       active, so a subsequent SET_CONFIGURATION allocated endpoint 1 over the
       still-live EP0 buffers. The allocator now re-reserves the EP0 IN/OUT
       buffers immediately after the reset; the bus-reset path is unchanged
       (github PR #85).
```

Verdict: advisory-LGTM (conditioned on the rebase + the changelog entry above landing with it). A human maintainer makes the merge decision.

gdisirio (approved):

PR #95, opened 2026-07-19
RP2: correct ROM region size and enable core-local BSS annotations

emolitor:
## Summary

Related fixes to the RP2 memory setup (review item 37):

- **ROM region size**: the `flash0` region in `RP2350_FLASH.ld`/`RP2350_RAM.ld` claimed 16k; the RP2350 boot ROM is 32k. Descriptive only — nothing is placed there — but the annotation was wrong.
- **Core-local BSS enablement**: `chmem.h` only activates the per-core `CH_MEM_LOCAL_BSS(n)`/`CH_MEM_LOCAL_COHERENT_BSS(n)` machinery when the *unsuffixed* `PORT_MEM_LOCAL_BSS`/`PORT_MEM_LOCAL_COHERENT_BSS` names are defined (it pastes the core suffix before expansion). All three RP2 SMP port headers (ARMv6-M, ARMv8-M-ML-ALT, and RISCV-HAZARD3) defined only the suffixed `...BSS0`/`...BSS1` variants, leaving the whole feature silently dead. Empty unsuffixed marker macros now switch it on; annotated variables land in `.ram4_clear.core0`/`.ram5_clear.core1`. The RT kernel immediately uses this for its per-core instance data (~1.8 KiB per core), which is the point of the feature — and what forces the next item.
- **Default stack sizes** (deliberate behavior change): `USE_EXCEPTIONS_STACKSIZE`/`USE_PROCESS_STACKSIZE` defaults become 0x400+0x400 in the RP2040, RP2350 and RP2350-RISCV startup makefiles (previously 0x800+0x800 on ARM, 0x400+0x800 on Hazard3; the RISCV-BLINK demo's own override is aligned too). The linker scripts place the per-core stacks in the 4 KiB scratch banks (ram4/SCRATCH_X for core 0, ram5/SCRATCH_Y for core 1); the old defaults left too little room for the newly-enabled core-local data — the ARM banks were completely full, and Hazard3's 1 KiB of headroom overflowed by 752 bytes the moment the feature came alive. 0x400+0x400 leaves half of each bank for annotated data and matches the common ChibiOS default on other ports.

**Impact note**: projects setting their own stack sizes are unaffected (`?=` defaults). A project silently relying on more than 1 KiB of default IRQ or main-thread stack now overflows at runtime instead of fitting; debug builds with stack checking catch this. Documented in the port headers: annotated data competes with stack space, and an overflow of annotated data surfaces as a link-time region error.

## Validation

- Map-file probe: `CH_MEM_LOCAL_BSS(0)`/`(1)` and coherent variants placed in ram4/ram5 as expected on RP2040 and RP2350 ARM builds.
- Hazard3: RISCV-BLINK links with each bank holding 0x400+0x400 stacks plus 0x6F0 of core-local kernel instance data (272 bytes spare); flashed to a Pico 2 in RISC-V mode and observed running (timer advancing, kernel scheduling) before restoring ARM firmware.
- Demo smoke runs on Pico 2 hardware with the new defaults; the full RP2350 remediation integration build (19 RP targets) also ran on hardware with these values.
- Reviewed by CodeRabbit and Codex, both clean; the automated PR review's should-fix (Hazard3 coverage) is addressed in d81018df66.

### DIFF
--- a/demos/RP/RT-RP2350-RISCV-BLINK/Makefile
+++ b/demos/RP/RT-RP2350-RISCV-BLINK/Makefile
@@ -54,7 +54,7 @@ endif
 # Stack size to be allocated to the RISC-V process stack. This stack is
 # the stack used by the main() thread.
 ifeq ($(USE_PROCESS_STACKSIZE),)
-  USE_PROCESS_STACKSIZE = 0x800
+  USE_PROCESS_STACKSIZE = 0x400
 endif
 
 # Stack size to the allocated to the RISC-V main/exceptions stack. This
--- a/os/common/ports/ARMv6-M/smp/rp2/chcoresmp.h
+++ b/os/common/ports/ARMv6-M/smp/rp2/chcoresmp.h
@@ -60,6 +60,18 @@
 #define PORT_SPINLOCK_NUMBER            31
 #endif
 
+/* Note: the ram4/ram5 scratch banks used by the following sections also
+   hold the per-core default stacks, annotated data competes with stack
+   space and an overflow surfaces as a link-time region error.*/
+
+/**
+ * @brief   Marker enabling suffixed @p PORT_MEM_LOCAL_COHERENT_BSSn
+ *          selection in chmem.h.
+ */
+#if !defined(PORT_MEM_LOCAL_COHERENT_BSS)
+#define PORT_MEM_LOCAL_COHERENT_BSS     /* Enables suffixed PORT_MEM_LOCAL_COHERENT_BSSn selection in chmem.h.*/
+#endif
+
 /**
  * @brief   Preferential local and coherent BSS section for core zero.
  */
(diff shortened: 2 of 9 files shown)


emolitor:
Actioned in d81018df66 — and the finding was better than it looked: adding the markers alone does not just enable an unused feature, the RT kernel immediately places its per-core instance data (~1.8 KiB per core) through `CH_MEM_LOCAL_BSS()`, and the Hazard3 defaults (0x400 exceptions + 0x800 process) left only 1 KiB per scratch bank — `ram4`/`ram5` overflowed by 752 bytes at link. The Hazard3 startup defaults and the RISCV-BLINK demo are therefore aligned to the same 0x400+0x400 split as the ARM ports in this PR.

Verified: RISCV-BLINK links with each bank holding 0x400+0x400 stacks plus 0x6F0 of core-local kernel data (272 bytes spare), and the firmware was flashed to a Pico 2 in RISC-V mode and observed running (timer advancing, kernel scheduling) before the board was restored to ARM firmware.

PR #96, opened 2026-07-19
RP SPIv1: fail loudly on DMA channel allocation failure

emolitor:
## Summary

`spi_lld_start()` ignored DMA channel allocation results (review item 23): a NULL rx channel was dereferenced later, and a NULL tx channel obtained after a successful rx allocation leaked the rx channel. Now:

- rx allocation failure returns cleanly;
- tx allocation failure frees the already-acquired rx channel first;
- both paths halt loudly in all build types rather than continuing into a corrupted configuration — the `osalSysHalt()` sites carry comments stating that effective contract, so the halts read as deliberate rather than as error handling to be "improved" into silent failure.

A compile-time `#error` rejects configurations that select the same DMA channel for RX and TX (except the ANY sentinel).

*Scope note: the shared DMA/PIO IRQ-vector priority contract (review item 26) was split out of this PR and is preserved on `em-shared-irq-priority` for a separate MR.*

## Validation

- RP2040 and RP2350 build matrix clean.
- GDB fault-injection of NULL allocation results on the Pico 2 exercised both failure paths.
- EFL-MFS and demo regression runs on hardware.
- Reviewed by CodeRabbit (clean) and Codex (one documentation finding, addressed in the halt-contract commit).

### DIFF
--- a/os/hal/ports/RP/LLD/SPIv1/hal_spi_lld.c
+++ b/os/hal/ports/RP/LLD/SPIv1/hal_spi_lld.c
@@ -171,7 +171,10 @@ void spi_lld_start(SPIDriver *spip) {
         /* A fixed-channel allocation failure is a static configuration
            conflict, continuing would dereference NULL later.*/
         osalDbgAssert(false, "no RX DMA channel");
-        osalSysHalt("SPI DMA alloc");
+        /* Formally an API-class function, every in-tree OSAL implements it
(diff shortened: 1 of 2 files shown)


emolitor:
Post-merge note on the stage-B should-fix: `python3 tools/style/stylecheck.py os/hal/ports/RP/LLD/SPIv1/hal_spi_lld.c` passes clean on the merged file, so the indentation finding does not reproduce with the project's own style gate — no follow-up planned.

PR #97, opened 2026-07-19
RP GPIOv1: full-port write atomicity and event vector core affinity

emolitor:
## Summary

Two PAL fixes on the RP2 port (review items 14 and 28), plus test coverage:

- **`pal_lld_writeport()` clobbered shared SIO latch bits**: the full-port write stored the value straight into `GPIO_OUT`, wiping the QSPI/USB-DPRAM latch bits that share the register's upper half on RP2350. It now performs a single atomic `GPIO_OUT_XOR` write of `(current ^ value)` masked to the port's valid GPIO lines, leaving non-GPIO latch bits untouched and staying glitch-free for the written pins (no momentary wrong level, unlike a set/clear pair).
  - *Documented limitation (from review)*: the read-XOR-write pair is not linearizable against concurrent same-port output updates from the other core — concurrent full-port writes must be serialized by the application. The set/clear/toggle line and group APIs remain safe for concurrent use among themselves. This contract is stated at the macro.
- **PAL event vector enabled on the wrong core**: the PAL event ISR reads the per-core `IO_BANK0` PROC block selected by `RP_PAL_EVENT_CORE_AFFINITY`, but `_pal_lld_init()` enabled the NVIC vector unconditionally on whichever core ran `halInit()`. With affinity 1, core 1's vector was never enabled (events never served) while core 0 held a vector for statuses it never sees. The vector is now enabled only when `SIO->CPUID` matches the affinity, and a new `palRPEnableEventsIrqX()` helper (asserting it runs on the affinity core) covers the case where the affinity core is not the one running `halInit()` — documented at the affinity setting.
- **Test coverage**: the multi/PAL test gains a boot self-check that seeds an inert non-GPIO latch bit, performs full-port writes, and verifies the latch half is preserved — the seed makes the check genuinely fail against the old behavior instead of passing trivially on all-zero latch bits.

## Validation

- On a Pico 2: latch-preservation self-check passes (and was confirmed to fail against the unfixed `pal_lld_writeport`); flash/UART remain alive across the write patterns (the QSPI latch bits genuinely matter there).
- RP2040 and RP2350 build matrix clean.
- Reviewed by CodeRabbit and Codex; both reviewers' findings on the concurrency contract are addressed in the two documentation commits.

### DIFF
--- a/os/hal/ports/RP/LLD/GPIOv1/hal_pal_lld.c
+++ b/os/hal/ports/RP/LLD/GPIOv1/hal_pal_lld.c
@@ -150,7 +150,14 @@ void _pal_lld_init(void) {
     _pal_init_event(i);
   }
 
-  nvicEnableVector(RP_IO_IRQ_BANK0_NUMBER, RP_IO_IRQ_BANK0_PRIORITY);
+  /* The ISR reads the interrupt statuses of the per-core PROC block
+     selected by RP_PAL_EVENT_CORE_AFFINITY and the NVIC is a per-core
+     resource, so the vector can only be enabled here when halInit() is
+     running on the affinity core. Otherwise the application must call
+     palRPEnableEventsIrqX() from the affinity core.*/
+  if (SIO->CPUID == (uint32_t)RP_PAL_EVENT_CORE_AFFINITY) {
+    nvicEnableVector(RP_IO_IRQ_BANK0_NUMBER, RP_IO_IRQ_BANK0_PRIORITY);
+  }
   #endif
 }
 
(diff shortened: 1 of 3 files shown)


emolitor:
Both findings actioned in the new commit:

- **Blocking (latch restore)**: the self-check now captures the full `GPIO_HI_OUT` upper half before seeding and restores it exactly at the end with an atomic masked XOR (same idiom as the driver fix), instead of blind-clearing the seeded bit — a bit already owned by earlier firmware survives the test on both success and failure paths (the restore runs before the failure signalling loop).
- **Should-fix (doc wording)**: the concurrency note now names set, clear *and toggle* consistently in both sentences.

Re-validated on a Pico 2: firmware runs through the self-check into normal demo operation (halted PC resolves to the idle thread, not the failure loop) and `GPIO_HI_OUT`'s shared half reads back exactly as before the test.

PR #101, opened 2026-07-19
RP ADCv1: report DMA allocation failure from adcStart

emolitor:
## Summary

`adc_lld_start()` ignored a failed DMA channel allocation (review item 13): the driver reported `ADC_READY` while holding no channel, and the next conversion dereferenced NULL. The driver now adopts the enhanced LLD API (`ADC_LLD_ENHANCED_API`): `adc_lld_start()` returns `HAL_RET_NO_RESOURCE` on allocation failure and the generic layer leaves the driver in `ADC_STOP`, making `adcStart()`'s return value meaningful.

Includes `testhal/RP/RP2350/ADC-VALIDATION`, whose regression leg is itself hardened: if the expected-to-fail start ever succeeds again (the historical bug), the cleanup only calls `adcStop()` when a DMA channel is actually held — under the regression the state lies (`ADC_READY` with a NULL channel) and a blind stop would halt inside the free before the test could report.

## Validation

On a Pico 2 (9/9):
- All DMA channels exhausted → `adcStart()` fails cleanly, state remains `ADC_STOP`.
- Channels released → `adcStart()` succeeds; a synchronous temperature-sensor conversion returns plausible non-zero samples (sensor warm-up handled by the test).
- Reviewed by CodeRabbit (test state hygiene, fixed) and Codex (clean; verification round added the NULL-channel cleanup guard).

### DIFF
--- a/os/hal/ports/RP/LLD/ADCv1/hal_adc_lld.c
+++ b/os/hal/ports/RP/LLD/ADCv1/hal_adc_lld.c
@@ -191,10 +191,11 @@ void adc_lld_init(void) {
  * @brief   Configures and activates the ADC peripheral.
  *
  * @param[in] adcp      pointer to the @p ADCDriver object
+ * @return              The operation status.
  *
  * @notapi
  */
-void adc_lld_start(ADCDriver *adcp) {
+msg_t adc_lld_start(ADCDriver *adcp) {
 
   /* If in stopped state then enables the ADC and DMA clocks.*/
   if (adcp->state == ADC_STOP) {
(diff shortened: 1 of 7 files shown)


emolitor:
Both should-fix items actioned in the new commit: the label now says what `taken > 0` actually proves (with exhaustion demonstrated by the failing start that follows), and the regression-path recovery uses `adcObjectInit()` instead of writing the internal state field. Re-validated on a Pico 2: 9/9.

emolitor:
Actioned in 22ea3c0d7d: honest exhaustion-check label and adcObjectInit() recovery. Re-validated 9/9 on a Pico 2. (Resolving the review threads accordingly.)

PR #103, opened 2026-07-19
RP2350: clock tree validation, E12 guard, safe tick-generator setup

emolitor:
## Summary

Clock-tree correctness work on the RP2350 (review items 17, 34, 35 plus errata hardening), six commits:

- **PLL settings validated at compile time**: the configured XOSC/REFDIV/FBDIV/postdiv combination is checked on *achieved* values — reference divisibility, reference ≥ 5 MHz, FBDIV in 16–320, VCO consistency, and exact divisibility down to the target frequency — so an unachievable clock request is a `#error`, not a silently wrong frequency. Matching runtime asserts in `rp_pll.c` validate denominators before any division (review-round fix: validation ordered ahead of use).
- **RP2350-E12 guard**: `#error` unless clk_sys×10 ≥ clk_usb×11, per the erratum's required ratio.
- **XOSC startup**: conservative startup-delay default (×6) and a range check on the computed `STARTUP` value.
- **Integer-MHz XOSC required + safe tick-generator sequence**: the TICKS generators are reprogrammed as disable → wait `!RUNNING` → write `CYCLES` → enable, instead of changing `CYCLES` under a running generator; non-integer-MHz crystals are rejected at compile time since the tick generators assume a 1 µs base.
- **Dropped dummy clock management**: `HAL_LLD_USE_CLOCK_MANAGEMENT` advertised a no-op `halclkcfg_t`/switch machinery; removed on both RP2040 and RP2350, while keeping `hal_lld_get_clock_point()` (called unconditionally by the generic HAL).

## Validation

On a Pico 2:
- FC0 frequency counter reads clk_sys at exactly 150000 kHz after init; TICKS `CYCLES` readback matches the crystal (12).
- Negative compile matrix: 13 MHz and 12.288 MHz crystal configurations fail with the intended `#error`s.
- Demo and EFL-MFS regression runs clean; RP2040 and RP2350 build matrix clean.
- Reviewed by CodeRabbit (clean) and Codex (denominator validation ordering, fixed and re-verified).

### DIFF
--- a/os/hal/ports/RP/RP2040/hal_lld.h
+++ b/os/hal/ports/RP/RP2040/hal_lld.h
@@ -62,11 +62,6 @@
 #define RP_GPIO_IOCTRL_OUTOVER_Pos          8
 /** @} */
 
-/**
- * @brief   Dynamic clock supported.
- */
-#define HAL_LLD_USE_CLOCK_MANAGEMENT
-
 /*===========================================================================*/
 /* Driver pre-compile time settings.                                         */
 /*===========================================================================*/
@@ -205,15 +200,6 @@
 /* Driver data structures and types.                                         */
 /*===========================================================================*/
 
-#if defined(HAL_LLD_USE_CLOCK_MANAGEMENT) || defined(__DOXYGEN__)
-/**
- * @brief   Type of a clock configuration structure.
- */
-typedef struct {
-  uint32_t          dummy;
-} halclkcfg_t;
-#endif /* defined(HAL_LLD_USE_CLOCK_MANAGEMENT) */
-
 /*===========================================================================*/
 /* Driver macros.                                                            */
 /*===========================================================================*/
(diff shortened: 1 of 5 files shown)


emolitor:
Actioned in the new commit:

- **Blocking (masked diagnostics)**: confirmed — `#error` does not halt preprocessing, so with e.g. `RP_PLL_SYS_REFDIV=0` the later derived checks emitted "division by zero in #if" noise around the intended message. Both PLLs' checks are now single `#elif` chains where a derived expression is only evaluated after every parameter it divides by passed its own range check (this also covers the postdiv-product divisibility check, which had the same hazard). Verified: a `-DRP_PLL_SYS_REFDIV=0` build now produces exactly one diagnostic, the intended range `#error`, with zero division-by-zero noise.
- **Style (mid-block declaration)**: `cycles` moved to the start of `rp_clock_init()`.

**Declined — E12 `ULL` multipliers**: C99 §6.10.1 requires `#if` arithmetic to be evaluated in `intmax_t`/`uintmax_t`, which is 64-bit on every toolchain this port supports, so `RP_CLK_SYS_FREQ * 10` cannot overflow in the preprocessor. Happy to add the suffixes if a pre-C99 toolchain ever becomes a target, but as-is they would suggest a hazard that does not exist.

emolitor:
New finding actioned: `rp_tick_start()` now asserts `index <= TICKS_RISCV` (the last of the six generators), matching the bounds checking style of the other helpers in the file. Demo re-run on a Pico 2 confirms boot through the guarded path with the tick generators live.

emolitor:
Actioned across 60cd666978 and 7dca315e8c: chained PLL check diagnostics (REFDIV=0 verified to produce a single clean #error), style fix, and the tick-generator bounds assert. The ULL-multiplier suggestion is declined: C99 mandates 64-bit (intmax_t) #if arithmetic, so the claimed overflow cannot occur on supported toolchains. (Resolving the review threads accordingly.)

emolitor:
Landing this change first then will update a new MR which reclocking based on the approach the STM32G4 uses.

PR #105, opened 2026-07-19
RP DMAv1: owner-derived channel free, RP2350 TRANS_COUNT mode handling

emolitor:
## Summary

DMA driver fixes on the RP2 port (review items 7 and 25):

- **Cross-core channel free used the wrong owner**: `dmaChannelFreeI()` keyed the per-core bookkeeping off `SIO->CPUID` — the core *executing* the free — rather than the core that allocated the channel. A channel allocated on core 0 and freed from core 1 corrupted both cores' allocation masks and could disable the wrong core's IRQ bookkeeping. Ownership is now derived from the recorded per-core allocation masks, and the cross-core free contract (caller must guarantee the channel's interrupt sources are quiesced; asserted via INTE0/INTE1) is documented and enforced.
- **RP2350 `TRANS_COUNT` is not a plain counter**: bits 31:28 select a mode (`NORMAL`/`TRIGGER_SELF`/`ENDLESS`). `dmaChannelSetCounterX()` passed the raw value through, so an oversized count silently programmed a mode. On RP2350 the count is now asserted to 28 bits and the mode field is forced to `NORMAL`; RP2040 is unchanged.
- Memory-ordering: a `DMB` pair ensures descriptor writes are published before the channel is triggered cross-core.

## Validation

`testhal/RP/RP2350/DMA-VALIDATION` on a Pico 2, 11/11:

- Cross-core free followed by alloc-all-16 succeeds; ownership masks verified.
- Mem-to-mem copies with `TRANS_COUNT` mode-nibble discriminators: counts 0x20000010 and 0xF0000010 both program mode NORMAL with the correct 28-bit count, verified via the `DBG_TCR` reload readback (the live down-counter reads 0 before trigger, so the debug reload register is the honest probe).
- Busy-channel disable path exercised.
- RP2040 and RP2350 build matrix clean; reviewed by CodeRabbit (2 fixed) and Codex (5 fixed) with hardware re-runs after each round.

### DIFF
--- a/os/hal/ports/RP/LLD/DMAv1/rp_dma.c
+++ b/os/hal/ports/RP/LLD/DMAv1/rp_dma.c
@@ -324,6 +324,17 @@ void dmaChannelFreeI(const rp_dma_channel_t *dmachp) {
                 "not allocated");
   osalDbgAssert(dmaChannelIsBusyX(dmachp) == false, "channel is busy");
 
+  /* A cross-core free cannot stop an interrupt handler already
+     executing on the owning core; the caller must have disabled the
+     channel's interrupt sources and quiesced its handler first. The
+     assertion catches the enabled-interrupt part of that contract.*/
+  osalDbgAssert(
+      (((dma.c0_allocated_mask & dmachp->chnmask) != 0U) ?
+       (SIO->CPUID == 0U) : (SIO->CPUID == 1U)) ||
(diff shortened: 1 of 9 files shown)


emolitor:
Actioned in 01740c5998: DMB on both sides of the cross-core free handoff (publish before flag, acquire before dependent allocations) and the TRANS_COUNT probes now use the CH_DBG structure member instead of the hard-coded address. Re-validated 11/11 on a Pico 2.

PR #106, opened 2026-07-19
RP TIMERv1: verify and retry system tick alarms set in the past

emolitor:
## Summary

The tickless system timer's `st_lld_set_alarm()` (review item 12) wrote the ALARM register and returned; if the target time had already passed by the time the write landed — interrupt-latency windows, debugger stalls, long critical sections — the alarm was missed entirely and the next timer event waited for a full 32-bit TIMERAWL wrap (~71 minutes) with the scheduler's virtual timers stalled.

The set-alarm path (and its `_n` variant) now verifies after arming: while the alarm is still ARMED and the target already lies in the past, it re-arms at `TIMERAWL + 2`. The ARMED check distinguishes "alarm already fired" (nothing to do) from "alarm armed for a time that will never come" (retry), which a naive time-comparison alone gets wrong on the fired-just-now edge.

## Validation

`testhal/RP/RP2350/ST-VALIDATION` on a Pico 2, 14/14:

- Two virtual timers with one deadline forced into the past by a busy-wait under `chSysLock`: the late timer fires within bounds (max observed latency 15 µs) instead of stalling.
- 10,000 randomized iterations of deadline-vs-latency races; a hardware watchdog backstops the run and watchdog resets are counted as failures (a stalled tickless timer would otherwise reset silently and restart the test).
- RP2040 and RP2350 build matrix clean (TIMER0 alias exists on both).
- Reviewed by CodeRabbit (2 test findings fixed) and Codex (ARMED-aware verify-and-retry replaced the original racy single re-arm), hardware re-validated after each round.

### DIFF
--- a/os/hal/ports/RP/LLD/TIMERv1/hal_st_lld.h
+++ b/os/hal/ports/RP/LLD/TIMERv1/hal_st_lld.h
@@ -220,11 +220,22 @@ __STATIC_INLINE void st_lld_set_alarm(systime_t abstime) {
      the check and the store (an interrupt or bus stall in the window
      recreates the ~71 min wrap wait), so the deadline is re-armed until
      either it is provably in the future or the compare has genuinely
-     fired (ARMED bit cleared by hardware) - a fired alarm must not be
-     re-armed into a duplicate event. */
-  while (((TIMER0->ARMED & 1U) != 0U) &&
-         ((int32_t)(target - TIMER0->TIMERAWL) <= 0)) {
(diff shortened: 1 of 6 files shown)


emolitor:
Actioned in 0c5a3b6fd3: both retry loops now sample the counter once per iteration and check ARMED immediately before re-arming, narrowing the fired-alarm window to a single read-to-store gap whose worst case is one spurious alarm interrupt (tolerated by the ST layer); the license header is added. Re-validated 14/14 with 10k randomized iterations, max latency 15 us.

PR #108, opened 2026-07-19
RP PIOv1: GPIOBASE support for GPIOs above 31

emolitor:
## Summary

Stacked on #107's base branch (`em-pio-fixes`) — this PR shows only the GPIOBASE commits and retargets to master automatically when that merges.

RP2350 PIO blocks address at most 32 GPIOs at a time; the `GPIOBASE` register (fixed as read-write in the register map by #92) selects whether a block sees GPIO0-31 or GPIO16-47 — required for the upper pins on RP2350B/QFN80 packages (review item 30):

- New registry capability `RP_PIO_HAS_GPIOBASE` (RP2350 TRUE, RP2040 FALSE).
- `pioSetGpioBase()` accepts 0 or 16, asserted against the capability and against changing the window while any state machine is allocated (the window shifts pin semantics under running programs).
- Pin-facing APIs gain `gpio < RP_GPIO_NUM_LINES` bounds checks; the `pioGpioToRel()` helper converts an absolute GPIO to the block-relative index, and the WAIT-instruction GPIO operand semantics (absolute vs relative) are documented.

## Validation

On a Pico 2 (RP2350A — the base window covers all its pins, so the high-window leg uses the LED):

- `GPIOBASE = 16` on PIO2, SET-pin program on relative pin 9 drives GPIO25 (LED), verified by `GPIO_IN` readback; raw register write to 0x50200168 silicon-probed during the register-map work.
- Full PIO-VALIDATION suite passes 17/17 with the GPIOBASE legs added; legs whose prerequisites are absent on this package are gated with explicit skips.
- Reviewed by CodeRabbit (1 fixed / 1 deferred to the package-registry PR, since merged as #102) and Codex (4 fixed including lifetime alignment with the base branch), hardware re-validated after each round.

### DIFF
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
@@ -668,10 +668,15 @@ void pioProgramUnload(const rp_pio_block_t *block,
  * @post    The block is taken out of reset and intentionally left out of
  *          reset so the setting persists across the following
  *          @p pioSmAllocI() calls (whose internal unreset is idempotent).
- * @note    Releasing the last state machine of the block puts the block
- *          back in reset which clears GPIOBASE to zero, the window must be
- *          configured again before the next allocation cycle if a
- *          non-default base is required.
+ * @note    The block is put back in reset (clearing GPIOBASE to zero)
+ *          only when it becomes fully idle: no state machines allocated
+ *          AND no program loaded. After such a full release the window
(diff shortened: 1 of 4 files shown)


emolitor:
Actioned in 3662037ced: pioGpioToRel() now checks the GPIOBASE read-back against the two windows the hardware implements (0 and 16) before the relative-pin arithmetic. Branch rebased onto the updated base; re-validated 17/17 on a Pico 2.

PR #139, opened 2026-07-22
RP2350 EFL: architecture-neutral interrupt masking

emolitor:
Stacked on #138 (retarget to master once that merges).

The RAM-resident flash operations in the RP2350 EFL driver mask
interrupts with the CMSIS PRIMASK intrinsics, which do not exist on the
Hazard3 RISC-V mode -- any RISC-V build enabling `HAL_USE_EFL` fails to
compile (found integrating the RP2350 RISC-V port into QMK, the first
RISC-V EFL consumer).

Replace the intrinsics with always-inline helpers that mask at the
architecture level: PRIMASK on ARM (behavior unchanged) and
`mstatus.MIE` on RISC-V. An OSAL/BASEPRI-based lock would not be
equivalent -- fast interrupts above the kernel priority ceiling must
also be deferred, because their handlers may execute from flash while
XIP is disabled. The helpers are forced inline so the masking code is
guaranteed to reside in the RAMFUNC callers.

The RISC-V mask/restore primitive lives in the Hazard3 port
(`hazard3_irq_disable_save()` / `hazard3_irq_restore()` in
`hazard3_irq.h`, alongside the existing `hazard3_irq_context_save/restore`
cookie idiom) rather than as raw `mstatus` CSR asm in the HAL driver, so
the driver's RISC-V arm is a named call symmetric with the ARM CMSIS
intrinsics. It is kept Hazard3-scoped, not promoted to the generic port
API which deliberately exposes only kernel-level locking; a full
interrupt disable is what this flash/XIP case specifically needs.

Verified the `__STATIC_FORCEINLINE` primitive inlines into the
RAM-resident flash functions (the `csrrc`/`csrs mstatus` land inside
`rp_flash_erase_full` at a `0x2000xxxx` SRAM address, not out-of-line in
XIP flash), so the RAMFUNC XIP-safety guarantee is preserved. Compile
coverage: ARM via the RP2350 EFL-SMP-LOCKOUT testhal, RISC-V via a QMK
Hazard3 build.

### DIFF
--- a/os/common/ports/RISCV-HAZARD3/hazard3_irq.h
+++ b/os/common/ports/RISCV-HAZARD3/hazard3_irq.h
@@ -313,6 +313,39 @@ __STATIC_FORCEINLINE void hazard3_irq_context_restore(uint32_t ctx) {
 }
 
 /**
+ * @brief   Disables all maskable interrupts, saving the previous state.
+ * @details Atomically reads mstatus while clearing mstatus.MIE. Unlike
+ *          @p port_disable() this returns the previous enable state so it
+ *          can be restored exactly, which is required by code that must
+ *          run with all interrupts masked but may be reached with them
+ *          already disabled. NMI and fault-class exceptions are not masked.
+ *
+ * @return  An opaque mask cookie to pass to @p hazard3_irq_restore(); the
+ *          caller must not interpret it.
+ */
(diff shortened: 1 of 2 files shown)


emolitor:
Reworking this into a change in hazard3_irq.h

emolitor:
Validated on RP2350 in both Hazard3 and ARM modes

PR #152, opened 2026-07-26
RP PIOv1: Add allocation masks and SM handle access

begeistert:
## Summary

This PR enhances the RP PIOv1 driver API by introducing new functions to query the allocation status of PIO state machines and their associated instruction memory. It also provides a convenient inline function, `pioGetSmHandleX`, to retrieve a descriptor handle for a specific state machine. These additions offer greater visibility and control over PIO resources, facilitating more robust and flexible application development by allowing users to check resource availability and easily access state machine configurations.

## Related

- Issue(s):
- Notes for reviewers: Reworked per review feedback — the state machine descriptor table (`__rp_pio_sms`) is now compile-time constant and exported `const`, so descriptors cannot be mutated by callers, handles are valid before `pioInit()` runs, and `pioGetSmHandleX()` calls with constant arguments fold to a constant descriptor. The accessor takes a block descriptor pointer, is `X`-suffixed/`@xclass`, and the allocation mask queries are documented as advisory snapshots (union of core 0 and core 1 allocations).

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

- [x] Build check passed (RP2040 and RP2350 PIO-VALIDATION targets; the POSIX simulator does not build this RP-specific LLD)

## Testing

- Any other tests run on a device? Yes — `PIO-VALIDATION` (extended with a
  test for the new functions) run on real hardware: Pico W (RP2040,
  50 asserts), Pico 2 ARM (57 asserts) and Pico 2 RISC-V (57 asserts),
  all passing. UART reports captured with a logic analyzer.

## Compatibility and Risk

- [ ] Risks/limitation are documented below

### DIFF
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
@@ -46,14 +46,35 @@ const rp_pio_block_t __rp_pio_blocks[RP_PIO_NUM_BLOCKS] = {
 #endif
 };
 
-/*===========================================================================*/
-/* Driver local variables and types.                                         */
-/*===========================================================================*/
-
 /**
  * @brief   PIO state machine descriptors.
  */
-static rp_pio_sm_t pio_sms[RP_PIO_NUM_BLOCKS][RP_PIO_NUM_STATE_MACHINES];
+const rp_pio_sm_t __rp_pio_sms[RP_PIO_NUM_BLOCKS][RP_PIO_NUM_STATE_MACHINES] = {
+  {
+    {&__rp_pio_blocks[0], 0U, 1U << 0},
+    {&__rp_pio_blocks[0], 1U, 1U << 1},
+    {&__rp_pio_blocks[0], 2U, 1U << 2},
+    {&__rp_pio_blocks[0], 3U, 1U << 3}
+  },
+  {
+    {&__rp_pio_blocks[1], 0U, 1U << 0},
+    {&__rp_pio_blocks[1], 1U, 1U << 1},
+    {&__rp_pio_blocks[1], 2U, 1U << 2},
+    {&__rp_pio_blocks[1], 3U, 1U << 3}
+  },
+#if RP_HAS_PIO2 == TRUE
+  {
+    {&__rp_pio_blocks[2], 0U, 1U << 0},
+    {&__rp_pio_blocks[2], 1U, 1U << 1},
+    {&__rp_pio_blocks[2], 2U, 1U << 2},
+    {&__rp_pio_blocks[2], 3U, 1U << 3}
+  },
+#endif
+};
+
+/*===========================================================================*/
+/* Driver local variables and types.                                         */
+/*===========================================================================*/
 
 /**
  * @brief   Global PIO-related data structures.
@@ -221,11 +242,6 @@ void pioInit(void) {
     pio.blocks[b].imem_allocated    = 0U;
     for (s = 0U; s < RP_PIO_NUM_STATE_MACHINES; s++) {
       pio.blocks[b].sm[s].func = NULL;
-
-      /* Initialize state machine descriptors.*/
-      pio_sms[b][s].block  = &__rp_pio_blocks[b];
-      pio_sms[b][s].smidx  = s;
-      pio_sms[b][s].smmask = 1U << s;
     }
   }
 }
@@ -328,7 +344,7 @@ const rp_pio_sm_t *pioSmAllocI(const rp_pio_block_t *block,
         pio.blocks[b].c1_allocated_mask |= smmask;
       }
 
-      return &pio_sms[b][i];
+      return &__rp_pio_sms[b][i];
     }
   }
 
@@ -686,6 +702,55 @@ void pioSmInit(const rp_pio_sm_t *smp, uint32_t initial_pc,
   pioSmRestartX(smp);
   pioSmClkdivRestartX(smp);
   pioSmSetPCX(smp, initial_pc);
+}
+
+/**
+ * @brief   Returns the allocation mask of the state machines in a block.
+ * @details The returned mask is the union of the core 0 and core 1
+ *          allocations, bit N representing state machine N.
+ * @note    The mask is a snapshot and is advisory only: another core can
+ *          allocate or free state machines right after the mask is taken.
+ *          @p pioSmAllocI() re-checks availability under the system lock,
+ *          so allocation remains safe regardless.
+ *
+ * @param[in] block     pointer to the PIO block descriptor
+ * @return              A bitmask where bit N represents state machine N.
+ *
+ * @api
+ */
+uint32_t pioGetSmAllocatedMask(const rp_pio_block_t *block) {
+  uint32_t mask;
+
+  osalDbgCheck(block != NULL);
+
+  osalSysLock();
+  mask = pio.blocks[block->pioidx].c0_allocated_mask |
+         pio.blocks[block->pioidx].c1_allocated_mask;
+  osalSysUnlock();
+
+  return mask;
+}
+
+/**
+ * @brief   Returns the allocation mask of the instruction memory in a block.
+ * @note    The mask is a snapshot and is advisory only, see
+ *          @p pioGetSmAllocatedMask().
+ *
+ * @param[in] block     pointer to the PIO block descriptor
+ * @return              A 32-bit mask representing the used instruction slots.
+ *
+ * @api
+ */
+uint32_t pioGetImemAllocatedMask(const rp_pio_block_t *block) {
+  uint32_t mask;
+
+  osalDbgCheck(block != NULL);
+
+  osalSysLock();
+  mask = pio.blocks[block->pioidx].imem_allocated;
+  osalSysUnlock();
+
+  return mask;
 }
 
 #if (RP_PIO_HAS_GPIOBASE == TRUE) || defined(__DOXYGEN__)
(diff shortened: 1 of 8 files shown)


emolitor:
Will take a look tomorrow!

emolitor (commented):
Thanks — the direction here is good. The two mask queries are useful and essentially right as-is (the locking matches how `pioSmAlloc()` guards the same state). The handle accessor needs rework before merge, though: as submitted it exposes a **mutable** global array in the public namespace, and the trade doesn't pay for itself. Below is an alternative shape that keeps everything this PR sets out to do.

## Main concern: the descriptor exposure

- `__rp_pio_blocks` is `const`; the newly exported `__rp_pio_sms` is not, so any including file can write to descriptors directly. Every API in this driver — including the new `pioGetSmHandle()` — returns `const rp_pio_sm_t *`, so writable exposure isn't needed by your own design; it's only a hazard.
- Because the array is non-const and filled at runtime by `pioInit()`, the inline can't constant-fold descriptor contents anyway: `smp->smidx`, `smp->smmask` and `smp->block` remain runtime loads. The inline only saves a call on setup-path code.
- A handle fetched before `pioInit()` runs is a zeroed descriptor (NULL `block`).

## Proposal: make the descriptor table compile-time `const`

`block`, `smidx` and `smmask` are all compile-time constants, and the driver never writes a descriptor after init (`pioSmAllocI()`/`pioSmFreeI()` only read them), so the table can be statically initialized and exported `const`:

```diff
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
@@ exported variables
 const rp_pio_block_t __rp_pio_blocks[RP_PIO_NUM_BLOCKS] = {
   {PIO0, 0U, RESETS_ALLREG_PIO0},
   {PIO1, 1U, RESETS_ALLREG_PIO1},
 #if RP_HAS_PIO2 == TRUE
   {PIO2, 2U, RESETS_ALLREG_PIO2},
 #endif
 };
 
+/**
+ * @brief   PIO state machine descriptors.
+ */
+const rp_pio_sm_t __rp_pio_sms[RP_PIO_NUM_BLOCKS][RP_PIO_NUM_STATE_MACHINES] = {
+  {
+    {&__rp_pio_blocks[0], 0U, 1U << 0},
+    {&__rp_pio_blocks[0], 1U, 1U << 1},
+    {&__rp_pio_blocks[0], 2U, 1U << 2},
+    {&__rp_pio_blocks[0], 3U, 1U << 3}
+  },
+  {
+    {&__rp_pio_blocks[1], 0U, 1U << 0},
+    {&__rp_pio_blocks[1], 1U, 1U << 1},
+    {&__rp_pio_blocks[1], 2U, 1U << 2},
+    {&__rp_pio_blocks[1], 3U, 1U << 3}
+  },
+#if RP_HAS_PIO2 == TRUE
+  {
+    {&__rp_pio_blocks[2], 0U, 1U << 0},
+    {&__rp_pio_blocks[2], 1U, 1U << 1},
+    {&__rp_pio_blocks[2], 2U, 1U << 2},
+    {&__rp_pio_blocks[2], 3U, 1U << 3}
+  },
+#endif
+};
+
 /*===========================================================================*/
 /* Driver local variables and types.                                         */
 /*===========================================================================*/
 
-/**
- * @brief   PIO state machine descriptors.
- */
-static rp_pio_sm_t pio_sms[RP_PIO_NUM_BLOCKS][RP_PIO_NUM_STATE_MACHINES];
-
 /**
  * @brief   Global PIO-related data structures.
  */
@@ pioInit()
   for (b = 0U; b < RP_PIO_NUM_BLOCKS; b++) {
     pio.blocks[b].c0_allocated_mask = 0U;
     pio.blocks[b].c1_allocated_mask = 0U;
     pio.blocks[b].imem_allocated    = 0U;
     for (s = 0U; s < RP_PIO_NUM_STATE_MACHINES; s++) {
       pio.blocks[b].sm[s].func = NULL;
-
-      /* Initialize state machine descriptors.*/
-      pio_sms[b][s].block  = &__rp_pio_blocks[b];
-      pio_sms[b][s].smidx  = s;
-      pio_sms[b][s].smmask = 1U << s;
     }
   }
 }
@@ pioSmAllocI()
-      return &pio_sms[b][i];
+      return &__rp_pio_sms[b][i];
```

```diff
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
@@ external declarations
 #if !defined(__DOXYGEN__)
 extern const rp_pio_block_t __rp_pio_blocks[RP_PIO_NUM_BLOCKS];
+extern const rp_pio_sm_t __rp_pio_sms[RP_PIO_NUM_BLOCKS]
+                                     [RP_PIO_NUM_STATE_MACHINES];
 #endif
@@ inline functions, before the config-builders group
+/**
+ * @brief   Returns the descriptor of a PIO state machine.
+ * @details The returned pointer is the same one @p pioSmAllocI() returns
+ *          for the state machine, so a handle can be recovered without
+ *          tracking the allocation-time pointer externally.
+ * @note    Holding a descriptor does not imply ownership: the state
+ *          machine must have been allocated with @p pioSmAlloc() or
+ *          @p pioSmAllocI() before it is operated on.
+ *
+ * @param[in] block     pointer to the PIO block descriptor
+ * @param[in] smidx     index of the state machine (0..3)
+ * @return              Pointer to the @p rp_pio_sm_t structure.
+ *
+ * @xclass
+ */
+__STATIC_INLINE const rp_pio_sm_t *pioGetSmHandleX(const rp_pio_block_t *block,
+                                                   uint32_t smidx) {
+
+  osalDbgCheck((block != NULL) && (smidx < RP_PIO_NUM_STATE_MACHINES));
+
+  return &__rp_pio_sms[block->pioidx][smidx];
+}
```

This keeps every capability of the PR (mask queries, handle-by-index, exported table) and adds:

- `const` exposure — no caller can corrupt descriptors;
- full constant folding: with `pioGetSmHandleX(RP_PIO0_BLOCK, 1)` the descriptor is a known constant, so downstream inlines like `pioSmEnableX()` compile to a single immediate store;
- ~96 B (RP2040) / 144 B (RP2350) move from `.bss` to flash, and `pioInit()` sheds the descriptor loop;
- handles are valid even before `halInit()`.

## Style/convention items

- The two new `.c` functions use 4-space indent; the file (and ChibiOS style) is 2-space.
- The `pioGetSmHandle` signature exceeds 80 columns.
- Every inline in this header is `X`-suffixed, and `@xclass` is the tag used for pure-data inlines (`@special` is used here for the hardware-touching SM inlines) — suggest `pioGetSmHandleX` + `@xclass`.
- The inline landed inside the "State machine configuration builders" `@name` group, which is documented as "pure data manipulation on a `rp_pio_sm_config_t`" — it belongs outside that group.
- Suggest taking `const rp_pio_block_t *block` instead of a raw index, consistent with the rest of the API (`RP_PIO0_BLOCK` etc. still fold to constants).
- Docs: `pioGetSmAllocatedMask()` should state the mask is the union of core 0 and core 1 allocations and is a snapshot (advisory only — `pioSmAllocI()` re-checks under lock). Also worth noting on the handle getter that holding a descriptor does not imply ownership.

## Checklist

- Style/build/test boxes are unchecked — please run the style check and compile at least one RP2040 and one RP2350 target (the latter exercises the `RP_HAS_PIO2` third table row).

Separately, once the API shape settles it would be good to extend `testhal/RP/*/PIO-VALIDATION` with asserts for the new functions (mask accounting including the cross-core union, imem bits on load/unload, and `pioGetSmHandleX()` returning pointer-equal to the `pioSmAlloc()` result). Happy to take that as a follow-up if you'd rather keep this PR focused.

begeistert:
Thanks for the thorough review @emolitor 😄 I adopted the proposal essentially
as-is, amended into the original commit and force-pushed, so the PR stays a
single focused change.

What changed relative to what you reviewed:

- `__rp_pio_sms` is now compile-time constant and exported `const`, with the
  static initializer including the third row under `RP_HAS_PIO2`. The
  descriptor loop is gone from `pioInit()`.
- The accessor is now `pioGetSmHandleX(const rp_pio_block_t *, uint32_t)`,
  tagged `@xclass`, moved out of the configuration-builders group, and the
  signature fits in 80 columns. Its docs carry the ownership note you
  suggested.
- The two mask queries are reindented to 2 spaces and documented as advisory
  snapshots: the union of core 0 and core 1 allocations, with `pioSmAllocI()`
  re-checking under lock.

Mechanical gates: stylecheck passes on both files, and both
`testhal/RP/RP2040/PIO-VALIDATION` and `testhal/RP/RP2350/PIO-VALIDATION`
build (the RP2350 one exercises the PIO2 table row). The checklist now says
`master`; the `main` wording was a leftover from the template.

For context on where this API is headed: the consumer is the nanoFramework
native layer (nanoframework/nf-interpreter#3449), which currently keeps its
own `g_AllocatedSMs[][]` table mirroring what the HAL already knows. With the
`const` table and `pioGetSmHandleX()` that bookkeeping can be deleted
entirely: handle recovery goes through `pioGetSmHandleX()` and the
claimed-check through `pioGetSmAllocatedMask()`. The advisory caveat doesn't
bite there because the CLR is the single allocator and runs single-threaded.
That migration will land on the nanoFramework side once this merges.

On extending PIO-VALIDATION: I took your suggestion directly in this PR.
Test 10 asserts the cross-core mask union (core 1 allocates while core 0
queries), the imem bits across load/unload, and pointer equality between
`pioGetSmHandleX()` and the `pioSmAlloc()` result. It ran on real hardware
on all three targets: Pico W (RP2040), and Pico 2 in both the ARM and
RISC-V builds, the latter two exercising the `RP_HAS_PIO2` table row.
Every assert passed.

emolitor (approved):
Thank you for quickly turning around the changes and adding the tests. Your contribution is appreciated!

PR #164, opened 2026-07-31
STM32U0xx: add EFL driver

emolitor:
Adds an embedded flash (EFL) low level driver for the STM32U0xx family, cloned from the STM32G0xx driver and adapted to the U0 flash controller.

- Single bank, 2K pages, 8 byte program lines.
- Size descriptors for all shipped U0 flash sizes (16K to 256K), selected at runtime from the flash size register (FLASHSIZE_BASE).
- The unconditional FLASH_SR_RDERR check from the G0 read path is dropped: the U0 flash status register has no read error flag.
- Dual bank paths compile out since the U0 CMSIS headers define neither FLASH_CR_MER2 nor FLASH_CR_BKER.

## Hardware validation

Validated on STM32U083 Rev A silicon (NUCLEO-U083RC, DEV_ID 0x489; identical flash controller and registry block to U073):

- **Geometry probe**: the runtime size-register match selected the 256 KiB descriptor (128 x 2 KiB sectors, single bank) on every init; no descriptor mismatch asserts.
- **Program path**: several hundred 8-byte line programs into the last sector, driven by a wear-leveling storage layer, each with immediate memory-mapped read-back verification. Programmed lines were additionally verified by external SWD readout and survive reset: the storage layer's log replay reads them back correctly on every boot.
- **Erase path**: exercised by storage consolidation once the 2 KiB log area filled: efl_lld_start_erase_sector + flashQueryErase polling + verify completed the sector erase in 22 ms, after which the consolidated image and its checksum were rewritten and validate on every subsequent init.
- **Lifecycle**: eflStart/eflStop cycled around every write session throughout (hundreds of cycles), no stuck states.
- **Errors**: no WRPERR/PROGERR/PGAERR decodes and no FASTERR/MISERR/SIZERR assertions observed at any point.

Independent of #165; no merge ordering constraint between the two.

### DIFF
--- a/os/hal/ports/STM32/STM32U0xx/hal_efl_lld.c
+++ b/os/hal/ports/STM32/STM32U0xx/hal_efl_lld.c
@@ -167,8 +167,10 @@ static inline void stm32_flash_lock(EFlashDriver *eflp) {
 
 static inline void stm32_flash_unlock(EFlashDriver *eflp) {
 
-  eflp->flash->KEYR |= FLASH_KEY1;
-  eflp->flash->KEYR |= FLASH_KEY2;
+  /* KEYR is write only, a read-modify-write could send a corrupted key
+     and lock the flash until the next reset.*/
+  eflp->flash->KEYR = FLASH_KEY1;
+  eflp->flash->KEYR = FLASH_KEY2;
 }
 
 static inline void stm32_flash_enable_pgm(EFlashDriver *eflp) {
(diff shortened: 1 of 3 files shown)


emolitor:
@gdisirio I needed this for a project last week. It is derived from the G0 which looked like the closest EFL hardware (maybe even the same silicon IP?). I did tidy a few things up and while it might be possible to have a common G0/L4/U0 EFL driver that isn't the pattern today so proposing adding it as a separate LLD.

gdisirio (approved):

PR #168, opened 2026-08-01
RP PIOv1: Support the RP2350 FIFO join modes

emolitor:
## Summary

The RP2350 adds `FJOIN_RX_GET` (SHIFTCTRL bit 14) and `FJOIN_RX_PUT` (bit 15), which disable the RX FIFO and repurpose its storage as a 4-entry register file with random access from the state machine and/or the processor. The driver modeled neither the bits nor the modes, so a consumer wanting them had to poke `cfgp->shiftctrl` internals directly (as the nanoFramework PIO binding, nanoframework/nf-interpreter#3449, does today — an encapsulation break that dangles if the config struct ever changes).

This PR:

- adds the two SHIFTCTRL bit definitions (RP2350-gated);
- extends `rp_pio_fifo_join_t` with `TXGET` (4), `TXPUT` (8) and `PUTGET` (12) — values deliberately matching the pico-sdk `pio_fifo_join` encoding, where enum bit 2 maps onto `FJOIN_RX_GET` and bit 3 onto `FJOIN_RX_PUT`;
- teaches `pioSmConfigSetFifoJoinX()` to program (and clear) all four FJOIN bits, with the debug check accepting exactly the six valid modes.

RP2040 behavior is unchanged: the new enum values and bits do not exist there, and the RP2040 builder body is untouched.

## Related

- Issue(s): —
- Notes for reviewers: naming follows pico-sdk ("get"/"put" name the state-machine side of the access); the enum doc spells this out plus the processor-side RXFx_PUTGETy window. Independent of #152 and #167.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

- [x] Build check passed (RP2040 and RP2350 PIO-VALIDATION targets; the POSIX simulator does not build this RP-specific LLD)

## Testing

- Any other tests run on a device? Yes — PIO-VALIDATION extended with Test 11 (FIFO join modes) exercising the new APIs, run on the bench at c0273bc938: RP2040 42 asserts (classic modes), RP2350 ARM 54 asserts (including the register-file modes), all passing. UART reports captured through the cross-wired CMSIS-DAP debug probes.

## Compatibility and Risk

- [ ] Risks/limitation are documented below


emolitor:
@begeistert worth you :eye: as this should simplify your nf-interpreter changes.

emolitor:
Local CodeRabbit CLI pass on this branch: 1 minor finding, declined (0 fixed / 1 declined). It claimed the RP2350 debug check accepts the undefined join value 3 via `(uint32_t)join <= RP_PIO_FIFO_JOIN_RX`; with `RP_PIO_FIFO_JOIN_RX = 2` that range check rejects 3 (3 <= 2 is false, and 3 matches none of the TXGET/TXPUT/PUTGET equalities), so the proposed explicit-equality rewrite is behaviorally identical. Keeping the `<=` form as it mirrors the pre-existing RP2040 branch of the same function.

begeistert:
It does simplify them, nicely! Today the binding clamps the managed enum with `join & 3` before calling `pioSmConfigSetFifoJoinX()`, precisely because the builder could not express the PUTGET modes, and you are right that reaching into `cfgp->shiftctrl` for the rest was an encapsulation break waiting to bite. With this PR the managed `PioFifoJoin` enum can mirror the pico-sdk values and pass straight through, the mask goes away, and no shiftctrl poking remains. I will update the binding once this merges.

Choosing the pico-sdk encoding was the right call in my opinion. I had been sketching the same extension with sequential enum values and your bit-mapped version is simply better, so I am glad this landed here first.

If testing a PUTGET-mode program on real hardware would be useful at any point, I am happy to run it.

emolitor:
Bench validation complete at c0273bc938. PIO-VALIDATION extended with Test 11: the classic modes produce the documented TX depths on both chips (4 unjoined, 8 joined — behavioral, via put-until-full on a disabled SM), and on the RP2350 each register-file mode programs exactly its FJOIN bit combination, keeps the TX FIFO at 4 entries, and clears cleanly on return to a classic mode. Results: **RP2040 42/42, RP2350 ARM 54/54**.

emolitor:
Round response (2 fixed / 0 declined) in 03a6b37a3d, addressing both "not ready" items: the FIFO-join builder branch is now gated `#if defined(RP2350) || defined(__DOXYGEN__)`, matching the guard on the enum values and SHIFTCTRL bits it references, so a documentation parse sees a consistent modes/implementation pair; and the test-number gap is the cross-PR reservation (10..13 across #167-#170), now documented inline in the suite header. No binary change on either chip (`__DOXYGEN__` is never defined in real builds); both targets rebuilt clean, hardware results at c0273bc938 stand. Local CodeRabbit delta pass on the test commit: no findings. Re-requesting review.

emolitor:
Round response (1 fixed / 1 declined) in 77a1988679.

**Fixed**: the enum note now references `PIO_SM_SHIFTCTRL_FJOIN_RX_GET` / `PIO_SM_SHIFTCTRL_FJOIN_RX_PUT` by their full public names — the short forms were indeed ambiguous and unlinkable.

**Declined** (final position, deferring further arbitration to the maintainer at merge time): renumbering Test 11 to 10 and dropping the coordination note. The four sibling PRs (#167-#170) insert tests at the same point in the same files; renumbering each to 10 guarantees four-way merge collisions **plus** three post-merge renumbering commits, strictly worse than the documented reservation. The review on #170 explicitly endorsed the document-the-reservation resolution now present in the header; this review asks for the opposite. The note is also not "stale" after an isolated merge — it explains exactly why a gap exists until the siblings land, and the last sibling's merge makes the numbering contiguous, at which point the note can be dropped in a trivial follow-up if desired. Re-requesting review.

emolitor:
Local CodeRabbit pass over this branch: 2 minor findings, 1 fixed / 1 declined. **Fixed** in the latest commit: the enum-level note now states the processor-access rule per mode (RXFx_PUTGETy is a processor path in TXGET/TXPUT only; PUTGET storage is SM-private). **Declined**: the claim that join value 3 passes the RP2350 debug check — `(uint32_t)join <= RP_PIO_FIFO_JOIN_RX` with `RX = 2` rejects 3, and 3 matches none of the TXGET/TXPUT/PUTGET equalities; this is the same misread CodeRabbit raised against the identical RP2040 branch idiom earlier.

emolitor:
Post-#152 conflict resolution and re-validation: master merged in (visible merge commit), this PR's test renumbered to **Test 12** (post-#152 reservation is 11..14 across #167-#170), sequenced after the allocation-masks test. Bench results on the merged tree at 27d471529a: **RP2040 53/53, RP2350 ARM 65/65**, all passing — this also exercises #152's test 10 and this PR's additions together on both chips. Branch is MERGEABLE again.

gdisirio (approved):

gdisirio (approved):

PR #172, opened 2026-08-02
I2Cv4: support devices without SMBus flags

emolitor:
The STM32U0xx I2C peripheral does not have SMBus support and its CMSIS headers do not define I2C_ISR_PECERR, I2C_ISR_TIMEOUT and I2C_ISR_ALERT, so hal_i2c_lld.c fails to compile for U0. Defining the missing flags as 0 makes the SMBus error decode paths compile out; no behaviour change for devices that have the flags.

Hardware validated on STM32U083 (NUCLEO-U083RC-hosted bring-up of a U073 target): I2C1 at 400 kHz (TIMINGR-configured, 56 MHz kernel clock) driving an IS31FL3741A LED controller — transactions ACK and the full register init sequence completes.

### DIFF
--- a/os/hal/ports/STM32/LLD/I2Cv4/hal_i2c_lld.c
+++ b/os/hal/ports/STM32/LLD/I2Cv4/hal_i2c_lld.c
@@ -60,8 +60,8 @@
 /*===========================================================================*/
 
 /* Devices without SMBus support (STM32U0xx) do not define the SMBus
-   related flags, defining them as zero makes the related error paths
-   compile out.*/
+   related flags; defining them as zero allows the related error paths
+   to be compiled out.*/
 #if !defined(I2C_ISR_PECERR)
 #define I2C_ISR_PECERR      0U
 #endif
--- a/os/hal/ports/STM32/LLD/I2Cv4/hal_i2c_lld.c
+++ b/os/hal/ports/STM32/LLD/I2Cv4/hal_i2c_lld.c
@@ -59,6 +59,21 @@
 /* Driver constants.                                                         */
 /*===========================================================================*/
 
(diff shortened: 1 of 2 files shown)


emolitor:
@gdisirio another minor change that came up this weekend bringing up a STM32U073 based keyboard for a personal project I've been working on.

gdisirio:
It is fun Copilot taking over the sheriff role lol. Thanks, I will look into this change ASAP.

chibios-sheriff (commented):

Scope: guard the SMBus-only ISR flags in the I2Cv4 error mask so the driver builds on devices without SMBus (STM32U0xx). Base `master`.

> not the full review yet.

Correct and confirmed with a repro. STM32U0xx uses I2Cv4 but its CMSIS headers don't define `I2C_ISR_PECERR`/`TIMEOUT`/`ALERT` (verified: absent on U0, present on U5), so `I2C_ERROR_MASK` fails to compile there. The `#if !defined(...) 0U` guards fix it and are no-ops where the flags exist.

Mechanical gates:
- Repro: building I2Cv4 on `stm32u083rc_nucleo64` (I2C1 enabled) fails on current `master` — `'I2C_ISR_TIMEOUT' undeclared` at `hal_i2c_lld.c:548` — and builds clean with this PR. ARM GCC 14.3.1.
- Style: changed lines clean. (`stylecheck.py` reports one lower-case comment at line 1037 — pre-existing, outside this PR's lines; not a finding here.)

Reviewer-owned changelog: this fixes a build break on released U0 I2Cv4, so it takes a `readme.txt` entry — I'll push it to this branch separately.

gdisirio (approved):

PR #191, opened 2026-08-07
STM32G0xx EFL: Add support 256K/512K devices

zvecr:
## Summary

`NUCLEO-G0B1RE` (and `xC` based boards) currently fail to configure a valid flash configuration due to the missing config

https://github.com/chibios-upstream/chibios/blob/fa13c03d8d8ac870553de0b92d651bbab4367361/os/hal/ports/STM32/STM32G0xx/hal_efl_lld.c#L205

## Related

- Issue(s):
- Notes for reviewers:
 
## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

## Testing

## Compatibility and Risk

### DIFF
--- a/os/hal/ports/STM32/STM32G0xx/hal_efl_lld.c
+++ b/os/hal/ports/STM32/STM32G0xx/hal_efl_lld.c
@@ -89,6 +89,38 @@ static const flash_descriptor_t efl_lld_size2[STM32_FLASH_NUMBER_OF_BANKS] = {
       }
 };
 
+/* The descriptor for 256K devices. */
+static const flash_descriptor_t efl_lld_size3[STM32_FLASH_NUMBER_OF_BANKS] = {
+      { /* Bank 1. */
+       .attributes        = FLASH_ATTR_ERASED_IS_ONE |
+                            FLASH_ATTR_MEMORY_MAPPED |
+                            FLASH_ATTR_ECC_CAPABLE   |
+                            FLASH_ATTR_ECC_ZERO_LINE_CAPABLE,
+       .page_size         = STM32_FLASH_LINE_SIZE,
+       .sectors_count     = STM32_FLASH_SECTORS_TOTAL_256K,
+       .sectors           = NULL,
+       .sectors_size      = STM32_FLASH_SECTOR_SIZE_256K,
+       .address           = (uint8_t *)FLASH_BASE,
+       .size              = STM32_FLASH_SIZE_256K * STM32_FLASH_SIZE_SCALE
+      }
+};
+
+/* The descriptor for 512K devices. */
+static const flash_descriptor_t efl_lld_size4[STM32_FLASH_NUMBER_OF_BANKS] = {
+      { /* Bank 1. */
+       .attributes        = FLASH_ATTR_ERASED_IS_ONE |
+                            FLASH_ATTR_MEMORY_MAPPED |
+                            FLASH_ATTR_ECC_CAPABLE   |
+                            FLASH_ATTR_ECC_ZERO_LINE_CAPABLE,
+       .page_size         = STM32_FLASH_LINE_SIZE,
+       .sectors_count     = STM32_FLASH_SECTORS_TOTAL_512K,
+       .sectors           = NULL,
+       .sectors_size      = STM32_FLASH_SECTOR_SIZE_512K,
+       .address           = (uint8_t *)FLASH_BASE,
+       .size              = STM32_FLASH_SIZE_512K * STM32_FLASH_SIZE_SCALE
+      }
+};
+
 /* Table describing possible flash sizes and descriptors for this device. */
 static const efl_lld_size_t efl_lld_flash_sizes[] = {
       {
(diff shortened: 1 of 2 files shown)


chibios-sheriff (commented):

Scope: add STM32G0xx EFL flash descriptors for 256K and 512K devices. Base `master`.
Stage reached: C (deep) — build/structure/style clean; a hardware-correctness finding on the 512K case.

> not the full review yet.

Thanks for this, and welcome. The mechanics are clean: the descriptors follow the existing `size1`/`size2` pattern exactly, the geometry math is right (256K = 128 × 2K, 512K = 256 × 2K, all 2K pages), and `stylecheck.py` is clean on both files.

**Blocking (Stage C — hardware/registers): the 512K case doesn't work on a factory-default G0B1.**

The 512K G0 parts (G0B1/G0C1) are physically **dual-bank**, and ST ships them in dual-bank mode: `FLASH_OPTR` reset value is `0xFFFF FEAA`, so bit 21 `DUAL_BANK = 1` (RM0444 §3.7.8). In that mode (RM0444 Table 11) the flash is **not** a contiguous 256-page region — Bank 1 is pages 0–127 and Bank 2 is pages **256–383**, addressed via `FLASH_CR.BKER` + a bank-relative `PNB` (0–127).

This driver is single-bank-only by design — `stm32_flash_dual_bank()` always returns `false` (and `#error`s if `STM32_FLASH_NUMBER_OF_BANKS > 1`), so the `BKER` branch in `efl_lld_start_erase_sector()` is dead. With the new 512K descriptor, erasing sectors 0–127 works (Bank 1), but sectors **128–255** get `PNB = sector`, `BKER = 0` — Bank 1 has no such pages, so the erase raises `FLASH_SR_SIZERR` (which the driver turns into an `osalDbgAssert`). Net: the upper 256 KB is unusable on a factory G0B1.

So the 512K support only holds if the device has been reconfigured to single-bank (`OPTR.DUAL_BANK = 0`, which itself needs a mass-erase). Could you say how you validated on the NUCLEO-G0B1RE — was the option byte set to single-bank, or was only the low 256 KB exercised?

The same question applies to **256K**: those parts are also dual-bank-capable (RM0444 Table 10), so if a given 256K part defaults to dual-bank the 128-page linear model has the same issue for its upper half.

**Root cause / direction (maintainer call):** the single-bank-only limitation is pre-existing in this driver, not something you introduced — so this isn't a "you got it wrong" note. The clean resolutions are either (a) document + guard that G0 512K/256K EFL requires `OPTR.DUAL_BANK = 0`, or (b) make the driver genuinely dual-bank aware (wire up the dead `BKER` path and a real `stm32_flash_dual_bank()`). @gdisirio may want to weigh in on which way to go.

Mechanical gates: `stylecheck.py` clean on both files. Build: the change is data-only (const descriptors) and matches the existing descriptor pattern; there's no in-tree G0 EFL target to link against, so I did not run a device build.

Verdict: comment — but with a **blocking** Stage-C finding: the 512K descriptor mis-addresses the upper bank on a factory-default (dual-bank) G0B1, so it isn't safe to merge as-is for the 512K case. I'm leaving the fix direction to the maintainer rather than formally blocking a first PR. Happy to re-review once the dual-bank handling (or an explicit single-bank requirement) is settled.

gdisirio:
@chibios-sheriff If there are no other blocking point let's document the limitation. Could you propose a driver enhancement for dual-bank support?

chibios-sheriff:
Confirmed — the dual-bank point was the only blocking item; the rest (geometry, descriptor pattern, style) is clean. Documenting the single-bank limitation is a fine resolution.

Proposed note to add near the size macros in `hal_efl_lld.h` (and the change is then good to go):

```c
/*
 * NOTE: the 256K and 512K descriptors assume single-bank operation
 *       (FLASH_OPTR.DUAL_BANK = 0). STM32G0B1/G0C1 (and dual-bank 256K
 *       parts) ship with DUAL_BANK = 1, where the flash is split into two
 *       banks with non-contiguous page numbering (RM0444 Table 11); this
 *       driver does not support that mode. On such devices the option byte
 *       must be set to single-bank before the upper flash can be erased or
 *       programmed through this driver.
 */
```

@zvecr — happy to push this to your branch (maintainer-access is on), or add it yourself, whichever you prefer. Either way, could you confirm how you validated on the NUCLEO-G0B1RE — single-bank option byte, or lower-256K only? Just so we know the documented path is exercised.

A proper dual-bank enhancement is being drafted separately per @gdisirio's request — that's out of scope for this PR, which stays the single-bank descriptor addition.

gdisirio (approved):

gdisirio (approved):

chibios-sheriff:
Pushed `0e8c04b944` at @gdisirio's request — the single-bank limitation note in `hal_efl_lld.h` (documentation only, stylecheck clean). Dual-bank support is now tracked in #192 as a follow-up enhancement.

@zvecr, that resolves the one finding — the PR is good from my side. Shout if you'd have worded the note differently.

zvecr:
`NUCLEO-G0B1RE` was tested with both the single bank, and dual bank option byte configuration.

The testing performed was manual, exercising the EFL driver VIA the EEPROM emulation code within `qmk_firmware`.

A user on Discord has also confirmed this works on a `G0B1CC` part.

PR #197, opened 2026-08-09
STM32: harden RTC clock validation

gdisirio:
## Summary

- make RTC compile-time checks use static frequency exports when dynamic clock points are runtime expressions
- add static RTC and PCLK1 frequency exports for the STM32C0, STM32U0, and STM32U3 HAL/XHAL ports
- enforce the required PCLK1 >= 7 x RTC relationship during dynamic clock switches
- validate the selected and enabled RTC clock source during STM32U5 clock switches
- mask the expected backup-domain RTC selector before comparing it, so RTCEN cannot force an unnecessary backup-domain reset

## Root cause

On dynamic-clock ports, STM32_RTCCLK can expand to a runtime clock-point lookup. Using that expression in a preprocessor condition breaks HSE/32 RTC configurations. Compile-time checks also cover only the default clock tree, so alternate runtime modes could bypass the APB/RTC ratio requirement. Generated clock-tree compatibility values can additionally carry RTCEN alongside RTCSEL, making the backup-domain selector comparison permanently unequal.

## Impact

Dynamic RTC configurations now compile on affected STM32C0, STM32U0, and STM32U3 targets, and invalid runtime APB/RTC combinations are rejected by the clock-tree validator. HAL and XHAL mirrors are kept aligned.

STM32C5 is intentionally excluded because that port is not complete enough to validate this change.

## Validation

- HAL STM32C0, STM32U0, and STM32U3 dynamic HSE/32 RTC builds
- XHAL STM32C0 dynamic HSE/32 RTC build
- XHAL STM32U0 and STM32U3 baseline builds
- STM32G4_TEST and STM32U3_TEST RTC-enabled clock-tree builds
- STM32G4_TEST and STM32U3_TEST builds with tautological-comparison warnings promoted to errors
- STM32U5 HAL and XHAL validation builds
- XHAL STM32G4 baseline build
- affected mcuconf/xmcuconf updaters run sequentially
- `git diff --check`

### DIFF
--- a/os/hal/ports/STM32/LLD/RCCv1/stm32_bd.inc
+++ b/os/hal/ports/STM32/LLD/RCCv1/stm32_bd.inc
@@ -74,7 +74,8 @@ __STATIC_INLINE void bd_init(void) {
 __STATIC_INLINE void bd_reset(void) {
 
   /* Reset BKP domain if different clock source selected.*/
-  if ((RCC->BDCR & RCC_BDCR_RTCSEL_Msk) != STM32_RTCSEL) {
+  if ((RCC->BDCR & RCC_BDCR_RTCSEL_Msk) !=
+      (STM32_RTCSEL & RCC_BDCR_RTCSEL_Msk)) {
     /* Backup domain reset.*/
 #if defined(RCC_BDCR_BDRST)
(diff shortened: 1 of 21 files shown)


CInsights (approved):
Also affected STM32U5.

PR #199, opened 2026-08-09
XHAL: initial RP2040/RP2350 port with ST and PAL support

emolitor:
## Summary

First stage of the XHAL port for RP2040/RP2350 — the first non-STM32 (and first SMP-capable) XHAL silicon port.

- `os/xhal/ports/RP`: twin of the classic RP port converted to the XHAL environment (no OSAL, `ch*` primitives directly): platform init (clocks/PLL/XOSC/bootrom), TIMERv1 system-timer LLD (TIMER0, 4 alarms, `ST_LLD_MULTICORE_SUPPORT`), GPIOv1 PAL LLD (`PAL_NEW_INIT`, event support, core-affinity extension), for both RP2040 and RP2350 (ARM).
- `demos/RP/RT-XHAL-RP-MULTI`: multi-target demo (Pico + Pico 2), PAL blinker, tickless, single-core at this stage. SIO console, SMP second core and the remaining drivers land in the next stages (SIO/SPI/WDG+DMA → I2C/ADC/PWM/RTC → USB/EFL → Hazard3).
- Build manifest (`OSAL_REMOVAL_BUILDS.txt`) and `port/RP` labeler globs extended.

Fork provenance: master `0db31951d5`; per-file-group table plus intentional divergences documented in `os/xhal/ports/RP/README.md`. Review aid: `git diff --no-index os/hal/ports/RP/<x> os/xhal/ports/RP/<x>` shows only the conversion delta.

Intentional divergences from classic:
- `hal_lld_init()` calls `stBind()` unconditionally — classic gates it to free-running mode, leaving the periodic SysTick path configured by `st_lld_bind()` but never started (classic has the same latent issue; a classic-side fix can follow separately).
- ST/PAL IRQ priority knobs validated with `CH_IRQ_IS_VALID_KERNEL_PRIORITY` (handlers take kernel locks; classic had no checks).

## Depends on

- #198 (`stBindAlarmN` indexed-alarm assert) — not required for this single-core stage, but required before the Stage-2 SMP demo enables the second core with assertions on.

## Verification

- Build matrix, all clean: rp2040_pico + rp2350_pico2 × {tickless, periodic (`CH_CFG_ST_TIMEDELTA=0`), debug (`CH_DBG_ENABLE_CHECKS/ASSERTS/SYSTEM_STATE_CHECK`)}, plus `USE_SMART_BUILD=no` on both targets (compiles every shared XHAL driver source against the new port).
[...]

PR #204, opened 2026-08-09
OSLIB: honor the revised semaphore contract in guarded pools

emolitor:
## Summary

#200 changed the semaphore contract: `chSemWaitTimeoutS()` now rejects `TIME_IMMEDIATE` and directs non-wait acquisition to `chSemFastWaitI()`. `chGuardedPoolAllocTimeoutS()` still forwarded its caller's timeout unmodified, so the documented guarded-pool non-waiting allocation (`chGuardedPoolAllocTimeout(gmp, TIME_IMMEDIATE)`) now stops on the parameter check in any build with `CH_DBG_ENABLE_CHECKS` enabled. The OS library test suite hits this deterministically in test case 7.2 (guarded memory pool without waiting).

This adopts the revised contract in the guarded pool: `TIME_IMMEDIATE` performs the acquisition directly on the counter under the system lock (`chSemGetCounterI` + `chSemFastWaitI`); all other timeouts keep the semaphore wait path. A sweep of `os/oslib/src` shows this is the only call site forwarding caller timeouts into the semaphore wait API.

## How it was found

Running the OSLIB test suite on RP2350 silicon with `CH_DBG_ENABLE_CHECKS`+`ASSERTS` (first hardware validation pass of the XHAL RP port, #199/#201): the suite halts in `chSemWaitTimeoutS` from `chGuardedPoolAllocTimeout`. With this fix the full RT + OSLIB suites pass on the Pico 2 (95 cases, 0 failures).

## Verification

- `test/rt/testbuild` POSIX simulator target builds clean (CI's build check).
- RT + OSLIB suites: Final result SUCCESS for both on RP2350 hardware with checks and assertions enabled.
- stylecheck clean.

## Changelog

- OSLIB: fixed guarded-pool non-waiting allocation stopping on the semaphore TIME_IMMEDIATE check with parameter checks enabled.

### DIFF
--- a/os/oslib/src/chmempools.c
+++ b/os/oslib/src/chmempools.c
@@ -321,11 +321,22 @@ void chGuardedPoolLoadArray(guarded_memory_pool_t *gmp, void *p, size_t n) {
  */
 void *chGuardedPoolAllocTimeoutS(guarded_memory_pool_t *gmp,
                                  sysinterval_t timeout) {
-  msg_t msg;
 
-  msg = chSemWaitTimeoutS(&gmp->sem, timeout);
-  if (msg != MSG_OK) {
-    return NULL;
+  /* Non-wait acquisition is performed directly on the counter, the
+     semaphore wait API no longer accepts TIME_IMMEDIATE.*/
+  if (timeout == TIME_IMMEDIATE) {
+    if (chSemGetCounterI(&gmp->sem) <= (cnt_t)0) {
+      return NULL;
+    }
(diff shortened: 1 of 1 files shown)


gdisirio:
Thank you for identifying this compatibility issue and proposing a fix. We decided to restore the previous `TIME_IMMEDIATE` behavior in the RT semaphore API instead, both for compatibility with existing callers and for consistency with NIL. The semaphore-level change is in #206, so the guarded-pool workaround proposed here should no longer be necessary.

PR #207, opened 2026-08-09
XHAL: RP port SPI driver with testxhal targets

emolitor:
## Summary

Stage 2b of the RP2040/RP2350 XHAL port: the PL022 SPI driver, converted from the classic LLD to the XHAL contract, plus RP2040/RP2350 targets for `testxhal/SPI`.

- `msg_t` operation returns; `setcfg`/`selcfg` with a static default configuration (Motorola, mode 0, 8-bit, ~1 MHz from the peripheral clock: SCR 61 @125 MHz, SCR 74 @150 MHz, compile-time guarded); frame size enforced from the mandatory `mode` field (8/16-bit, larger rejected); `SSPCR0` reserved encodings rejected (FRF=3, out-of-mask bits); live reconfiguration quiesces the PL022 (BSY wait, SSE off) before touching CR0/CPSR.
- **New `spi_lld_stop_transfer`**: `SSPDMACR` quiesce → erratum-safe abort of both DMA channels (RP2040-E13/RP2350-E5 path in the shared service) → residual frames from the RX channel counter (RP2350 `TRANS_COUNT` MODE field masked) → FIFO drain while BSY → DMACR restore.
- **Terminal-event arbitration** (from adversarial review): a per-driver transfer generation with parity semantics makes completion, DMA error, and abort single-winner across cores — an in-flight completion on the other core arbitrates atomically against `spiStopTransferI`; a stale RX-completion dispatched from the same DMA-IRQ snapshot as a TX error loses the claim (plus hardware guards: RX EN clear, `TRANS_COUNT`==0); channel-index ordering no longer affects correctness, and inverted explicit channel assignments are additionally rejected at start (`HAL_RET_NO_RESOURCE`).
- Completion maps to the COMPLETE transition (per SPIv2 precedent); DMA errors keep classic's force-completion behavior through the error path, with `RP_SPI_DMA_ERROR_HOOK` fired only by the winning claim.
- `SPI_SUPPORTS_CIRCULAR FALSE`, `SPI_SUPPORTS_SLAVE_MODE FALSE` (classic parity; circular is a candidate follow-up). Select mode: `RP_SPI_SELECT_MODE` (LINE default / PAD) mirroring the STM32 knob.
- `testxhal/SPI`: both RP targets (SPI0 on GP16/17/18/19, CS preset high before pad-mode switch), shared `main.c` untouched; recipes registered in the OSAL-removal manifest.

## Review triage

- codex adversarial review: 3 MAJOR cross-core race findings — all fixed by the generation/claim design above; 1 MINOR (reserved SSPCR0 encodings) fixed; 1 MINOR (CS setup glitch) fixed; 1 coverage note (the shared testxhal/SPI `main.c` compiles out `spiStopTransfer` sites when circular/slave are unsupported, so the new abort path needs dedicated bench coverage — see validation notes below).
- coderabbit CLI: 1 finding (CS glitch, same as above) — fixed.
- Review byproduct for a separate DMAv1 look (inherited shape, classic has it too): `rp_dma` `serve_interrupt()` writes `CTRL_TRIG` error-clear bits unconditionally before the callback, which can clobber a channel re-armed cross-core from a stale INTS snapshot; the SPI claim guards make this benign for SPI (no false terminal event), but a SET/CLR-alias or busy-skip hardening in the service would remove the hang window.

## Verification

- Build matrix, all clean: testxhal/SPI rp2040+rp2350 × {default, +`CH_DBG_ENABLE_CHECKS/ASSERTS`}; demo unaffected (smart-build isolation confirmed, `USE_SMART_BUILD=no` leg compiles the shared SPI driver against the new LLD); untouched STM32 testxhal/SPI leg still green.
- OSAL lexical gate, per-file stylecheck, `git diff --check`: clean. Zero shared XHAL changes.
- On-hardware validation to follow on the bench (Pico/Pico 2 via the probe rig): transfer/abort residual behavior and loopback data checks; will be posted here before this leaves draft.

## Changelog

- XHAL: RP port gains the SPI driver; testxhal/SPI gains RP2040/RP2350 targets.

### DIFF
--- a/os/xhal/OSAL_REMOVAL_BUILDS.txt
+++ b/os/xhal/OSAL_REMOVAL_BUILDS.txt
@@ -39,6 +39,8 @@ testxhal/I2S/make/stm32g474re_nucleo64.make
 testxhal/MMC-SPI/make/stm32g474re_nucleo64.make
 testxhal/RTC/make/stm32g474re_nucleo64.make
 testxhal/RTC/make/stm32wl55jc_nucleo64.make
+testxhal/SPI/make/rp2040_pico.make
+testxhal/SPI/make/rp2350_pico2.make
 testxhal/SPI/make/stm32g474re_nucleo64.make
 testxhal/SPI/make/stm32wl55jc_nucleo64.make
 testxhal/TRNG/make/stm32g474re_nucleo64.make
--- a/os/xhal/ports/RP/LLD/SPIv1/driver.mk
+++ b/os/xhal/ports/RP/LLD/SPIv1/driver.mk
@@ -1,0 +1,9 @@
+ifeq ($(USE_SMART_BUILD),yes)
+ifneq ($(findstring HAL_USE_SPI TRUE,$(HALCONF)),)
+PLATFORMSRC += $(CHIBIOS)/os/xhal/ports/RP/LLD/SPIv1/hal_spi_lld.c
+endif
+else
+PLATFORMSRC += $(CHIBIOS)/os/xhal/ports/RP/LLD/SPIv1/hal_spi_lld.c
+endif
+
+PLATFORMINC += $(CHIBIOS)/os/xhal/ports/RP/LLD/SPIv1
--- a/os/xhal/ports/RP/LLD/SPIv1/hal_spi_lld.c
+++ b/os/xhal/ports/RP/LLD/SPIv1/hal_spi_lld.c
@@ -1,0 +1,763 @@
+/*
+    ChibiOS - Copyright (C) 2006-2026 Giovanni Di Sirio.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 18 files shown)


gdisirio (approved):

emolitor:
On-hardware validation completed on both chips (test firmware with `CH_DBG_ENABLE_CHECKS`+`ASSERTS`; both boards' probe firmware backed up and restored afterwards, byte-verified):

- **Loopback data verification, RP2040 and RP2350**: testxhal/SPI has no button on the Pico, so the button was driven through the debugger (IO_BANK0 `INOVER` force on GP22) to start the contender threads, and the PL022 loopback bit (`SSPCR1.LBM`) was enabled inside a breakpoint window around `spiExchange`. Result on both chips: `rxbuf` matches `txbuf` byte-for-byte over the 512-frame DMA exchange (previously 0xFF with the pad floating) — full TX→RX data path through the new driver verified.
- **Bus-contention soak, both chips**: 60+ s with the two contender threads cycling `drvLock`/`drvStart`/`drvSelectCfgX`/`spiSelectX`/`spiExchange` under checks+assertions — no assertion, no panic, clean idling between DMA-driven transfers.
- Not yet exercised on silicon: the `spi_lld_stop_transfer` abort path (compiled out of the shared testxhal main for non-circular/non-slave targets, as noted in the review triage) and the terminal-race windows themselves (require instrumented stress firmware). Suggest treating those as candidates for a dedicated RP validation app in a follow-up, per the codex coverage note.

Bench technique note for reviewers: the virtual-button INOVER trick means these targets can be driven fully unattended through all test phases in future bench runs.

PR #208, opened 2026-08-09
XHAL: RP port PWM and ADC drivers

emolitor:
## Summary

Stage 3a of the RP2040/RP2350 XHAL port: the PWM and ADC drivers, plus live driver exercise in the RT-XHAL-RP-MULTI demo.

**PWM** (`LLD/PWMv1`): per-slice drivers (`PWMD0..11`, classic numbering kept) on the XHAL contract. RP slices expose only the wrap interrupt, so `PWM_EVENT_PERIOD` is the sole supported notification — channel event bits assert in debug builds, are stripped from the enabled-events cache in release builds, and configurations carrying them are rejected (the classic per-channel compare callback machinery was verified dead code — no compare interrupt exists, nothing ever invoked it — and is dropped). The shared PWM block's lifecycle is managed by a software refcount (active + start-in-progress masks, all transitions under the cross-core system lock) with the shared wrap vector owned by the first-user core — the base driver drops the lock before physical start/stop, so `PWM->EN` alone cannot arbitrate first-user/last-user transitions safely on SMP. Initial configuration is validate-only until the shared block is acquired; live reconfiguration keeps the quiesce-then-apply path. Stale wrap latches are cleared before unmasking.

**ADC** (`LLD/ADCv1`): the XHAL group model — conversion groups live in the configuration, selected by index with bounds/consistency validation (per-variant channel range, round-robin popcount vs `num_channels`, mask-includes-first-channel). The DMA-paced FIFO engine is retained for linear conversions; **circular conversions use two chained DMA channels in ping-pong**, eliminating the CPU re-arm window entirely (the classic single-channel re-arm would have had only a FIFO-depth deadline, ~8–16 µs at 500 kS/s; the deadline is now one half-buffer period). Temperature-sensor bias is owned per group and dropped at conversion stop, including the error path. Default configuration samples the on-die temperature channel.

**Demo**: after the test suites, the LED fades under PWM with a period-event counter and the on-die temperature prints periodically on the console — live silicon exercise of both drivers' full paths.

## Review triage

- codex adversarial review: 1 BLOCKER (PWM shared-resource lifecycle races on SMP) and 4 MAJOR (config-before-start hardware touch, channel-event cache lie, ADC validation holes, circular sample-loss window) — all redesigned/fixed as described above; 4 MINOR fixed (TS_EN lifetime, DMA-channel knob validation, `pwmChangePeriod` cache/hardware desync via clamp-writeback, negative-temperature print).
- coderabbit CLI: 2 findings (unshifted round-robin validation, negative-temperature sign) — both fixed.

## Verification

- Build matrix, all clean: demo rp2040+rp2350 × {default, +`CH_DBG_ENABLE_CHECKS/ASSERTS`}, `USE_SMART_BUILD=no`, testxhal/SPI regression leg. OSAL lexical gate, per-file stylecheck, `git diff --check` clean. Zero shared XHAL changes.
- On-hardware validation to follow before this leaves draft: LED fade + wrap-counter + temperature telemetry on both chips via the probe rig (unattended, using the established virtual-button technique).

## Changelog

- XHAL: RP port gains the PWM and ADC drivers; RT-XHAL-RP-MULTI exercises them live.

### DIFF
--- a/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/portab.c
+++ b/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/portab.c
@@ -34,6 +34,54 @@
 /* Module exported variables.                                                */
 /*===========================================================================*/
 
+/**
+ * @brief   PWM configuration for the LED slice.
+ * @details The slice counts at 1MHz over a 1000 ticks period giving a
+ *          1kHz PWM frequency, the LED on GP25 is driven by the B
+ *          channel. Events are enabled at runtime by the application.
+ */
+const hal_pwm_config_t portab_pwm_config = {
+  .frequency      = 1000000U,
+  .period         = 1000U,
+  .enabled_events = 0U,
+  .channels       = {
+    {
+      .mode       = PWM_OUTPUT_DISABLED
+    },
+    {
+      .mode       = PWM_OUTPUT_ACTIVE_HIGH
+    }
+  },
+  .dummy          = 0U
+};
+
+/**
+ * @brief   Conversion groups of the portability ADC configuration.
+ * @details A single conversion group sampling the on-die temperature
+ *          sensor at the free-running rate, the sensor bias is enabled
+ *          by the driver for the duration of the conversion.
+ */
+static const adc_conversion_groups_t portab_adc_groups = {
+  .grpsnum        = 1U,
+  .grps           = {
+    {
+      .num_channels = 1U,
+      .channel      = ADC_CHANNEL_TEMPSENSOR,
+      .rrobin       = 0U,
+      .div          = 0U,
+      .ts_enabled   = true
+    }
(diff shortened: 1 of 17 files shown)


emolitor:
RP2350 hardware validation complete (single contiguous capture, `CH_DBG_ENABLE_CHECKS`+`ASSERTS`): RT and OSLIB suites both `Final result: SUCCESS`, followed by 55 cycles of live driver telemetry — on-die temperature stable at 28.0–28.5 °C through the ADC group/DMA path, and the PWM period-event counter advancing at 1000.3 wraps/s against the host clock (1 kHz fade configuration; every wrap event delivered). The RP2040 leg will run in a batched multi-stage bench pass (requires the probe-swap procedure) before this and the sibling stage drafts leave review.

gdisirio (approved):

Issue #205, opened 2026-08-09
ARMv8-M-ML-ALT SMP: system state checker fails on RP2350 dual-core (classic and XHAL)


emolitor:
## Summary

Any dual-core RP2350 build with `CH_DBG_SYSTEM_STATE_CHECK=TRUE` halts within seconds of SMP activity — including the unmodified classic `demos/RP/RT-RP2350-PICO2` — so this predates the XHAL RP port and is masked in normal use because the classic demos ship with `-DNDEBUG`.

## Observed failures (Pico 2, via CMSIS-DAP probe)

1. XHAL RP demo (RT suite over buffered SIO console): `chSysHalt("SV#8")` — `__dbg_check_enter_isr` on ISR entry (`Vector40`/`VectorC4`) over the idle thread with `ch0.dbg.lock_cnt == 1`, `isr_cnt == 0`.
2. Classic `RT-RP2350-PICO2` built with `CH_DBG_ENABLE_CHECKS/ASSERTS/SYSTEM_STATE_CHECK`: `chSysHalt("SV#4")` — `__dbg_check_lock` from a plain `chSemWait` in the blinker thread.

## Evidence gathered so far

- A hardware watchpoint on `ch0.dbg.lock_cnt` shows perfectly balanced lock/unlock pairs (e.g. `oqPutTimeout`), then a legitimate `chSysLock` in `test_execute_inner`, then `Vector40` (TIMER0 alarm 0, IRQ priority 2, IPR verified `0x20`) enters while the counter is still 1 — despite `CORTEX_BASEPRI_KERNEL = 0x10` supposedly masking it. BASEPRI read back `0x10` at the halt.
- The SIO FIFO IPI (`VectorA4`) is registered at `CORTEX_MINIMUM_PRIORITY` with proper prologue/epilogue, so it is not an unmasked-IPI problem.
- `chdebug.c` uses `currcore` per check, so the counters are per-instance by design.
- Single-core configurations of the same binaries run clean (the XHAL RP demo passes the full RT+OSLIB suites with `CHECKS`+`ASSERTS`; the state checker is the only failing level, and only with the second core started).

It is unclear yet whether the state checker is producing false positives on this port (dbg accounting vs the ALT port's BASEPRI/spinlock locking scheme under SMP) or whether it is correctly detecting critical sections that are genuinely permeable on dual-core RP2350 — the second possibility would matter well beyond debug builds.

[...]

PR #212, opened 2026-08-09
XHAL: RP port USB device driver

emolitor:
## Summary

Stage 4b of the RP2040/RP2350 XHAL port — the final ARM driver: the USB device controller under the XHAL contract, plus RP targets for `testxhal/USB_CDC`.

- The hardware layer is preserved byte-faithfully (`rp_usb.h` is an untouched copy; DPRAM allocator, double-buffering, RP2350 abort teardown, RP2040-E15 workaround verified identical). The conversion is confined to the contract layer.
- Per-endpoint hardware state (`next_pid`/buffer geometry) that the shared XHAL endpoint structures no longer carry moves into LLD-private endpoint contexts; the EP0 state union follows the STM32 template, with the safety argument (directional state fully initialized before use, SETUP reseeding both directions) verified in review.
- The classic requirement trio maps onto the new shared model: late address setting via the retained `USB_SET_ADDRESS_MODE` (consumed by the shared status-stage thread — early SETUP correctly aborts stale application), while the software status-stage/address-ACK selections are now documented no-ops under the unconditional shared implementation.
- SOF interest moves from the classic config-callback test to binder presence evaluated at bus reset (binders attach before connect in the documented lifecycle — review confirmed the post-enumeration attach hole is unreachable through the normal API); E15 builds keep the continuous enable for erratum timing.
- No port-side terminal arbitration was added: the controller has a single vector and the shared busy bitmaps + EP0 sequence checks serialize all completion paths — reliance documented in-source.
- `testxhal/USB_CDC` gains both RP targets with the classic USB knob values and the full debug trio enabled (state checker included — these are single-core targets, unaffected by #205).

## Review triage

- codex adversarial review: **accept, no blockers or majors**. One MINOR fixed (debug trio in the new test configs). INFO items: the SOF lifecycle note above, and the shared test descriptor's ST VID/PID (`0483:5740`) — pre-existing collateral affecting all USB_CDC targets, flagged for a separate decision on a project test identity.
- coderabbit CLI: pending (rate-limited after today's review volume) — will be run and triaged before this leaves draft.

## Verification

- Build matrix clean, zero warnings: testxhal/USB_CDC rp2040+rp2350 × {default, debug}, `USE_SMART_BUILD=no` demo leg (shared `hal_usb.c`/`hal_usb_cdc.c` against the new headers), untouched STM32 leg.
- OSAL gate, per-file stylecheck, `git diff --check`: clean. Zero shared XHAL changes.
- On-hardware CDC enumeration/echo/reset-stress (host_check.py pattern) queued for the batched bench pass.

## Changelog

- XHAL: RP port gains the USB device driver; testxhal/USB_CDC gains RP2040/RP2350 targets.

### DIFF
--- a/os/xhal/OSAL_REMOVAL_BUILDS.txt
+++ b/os/xhal/OSAL_REMOVAL_BUILDS.txt
@@ -46,6 +46,8 @@ testxhal/SPI/make/stm32wl55jc_nucleo64.make
 testxhal/TRNG/make/stm32g474re_nucleo64.make
 testxhal/TRNG/make/stm32h563zi_nucleo144.make
 testxhal/TRNG/make/stm32wl55jc_nucleo64.make
+testxhal/USB_CDC/make/rp2040_pico.make
+testxhal/USB_CDC/make/rp2350_pico2.make
 testxhal/USB_CDC/make/stm32g474re_nucleo64.make
 testxhal/WDG/make/rp2040_pico.make
 testxhal/WDG/make/rp2350_pico2.make
--- a/os/xhal/ports/RP/LLD/USBv1/driver.mk
+++ b/os/xhal/ports/RP/LLD/USBv1/driver.mk
@@ -1,0 +1,9 @@
+ifeq ($(USE_SMART_BUILD),yes)
+ifneq ($(findstring HAL_USE_USB TRUE,$(HALCONF)),)
+PLATFORMSRC += $(CHIBIOS)/os/xhal/ports/RP/LLD/USBv1/hal_usb_lld.c
+endif
+else
+PLATFORMSRC += $(CHIBIOS)/os/xhal/ports/RP/LLD/USBv1/hal_usb_lld.c
+endif
+
+PLATFORMINC += $(CHIBIOS)/os/xhal/ports/RP/LLD/USBv1
--- a/os/xhal/ports/RP/LLD/USBv1/hal_usb_lld.c
+++ b/os/xhal/ports/RP/LLD/USBv1/hal_usb_lld.c
@@ -1,0 +1,1111 @@
+/*
+    ChibiOS - Copyright (C) 2006-2026 Giovanni Di Sirio.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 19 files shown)


emolitor:
coderabbit CLI pass now complete (it was rate-limited when this PR opened): 3 MAJOR findings, all verified **identical in `os/hal/ports/RP/LLD/USBv1`** and therefore left untouched here under the fork-fidelity policy this campaign has followed throughout — they are classic-side candidates, not conversion defects:

1. `usb_isochronous_buffer_size()` does not normalize 896 to 1024 (it special-cases 384, 640, 768 and >1024), so an isochronous endpoint with a max size in 769..896 programs an invalid double-buffer offset. Classic `hal_usb_lld.c:142-153` is character-for-character the same.
2. The OUT completion path uses the hardware-reported length without clamping it to the remaining `rxsize`. In practice the programmed buffer length is already `min(rxsize, out_maxsize)`, so the controller should not be able to report more — but the defensive clamp is absent in both trees.
3. The buffer-arm writes length, flags and `AVAILABLE` in a single store. The RP datasheets call for setting `AVAILABLE` in a second write after the rest of the buffer control word has landed (with a short delay and a barrier), because the controller may sample the control word before the CPU's write completes. Classic arms both the IN and OUT buffers the same way.

Item 3 is the one I would flag for attention beyond this PR: it affects shipped classic RP USB firmware, not just this port. Happy to prepare a classic-side fix (with the XHAL twin mirrored) as a separate change if that is wanted — it did not seem right to slip a hardware-sequencing change into a conversion PR whose whole claim is byte-faithfulness to the classic hardware layer.

gdisirio (approved):

emolitor:
The classic-side fix for all three findings is now up as **#214**, and this branch carries the identical change so the hardware layer stays byte-faithful to it.

It is validated on silicon (Pico 2, second Pico as SWD probe, target USB to the host): the `host_check.py` echo leg passes with zero failures across 50 rounds × 10 packet-boundary sizes (~674 KB, device-side byte counter agreeing), and a re-enumeration stress of 15 full target resets — each forcing a USB disconnect, re-enumeration and complete endpoint re-arm, which is exactly the path the buffer-arm fix touches — passes 15/15 with the echo battery repeated after every cycle.

Worth noting for whoever picks up the descriptor question: the device enumerates on the host as "STM32 Virtual COM Port, 0483:5740" while running on a Raspberry Pi part, which is the identity issue from the earlier INFO finding observed directly rather than inferred.

PR #213, opened 2026-08-09
XHAL: RP port I2C driver with asynchronous abort

emolitor:
## Summary

Stage 3b of the RP2040/RP2350 XHAL port and the campaign's largest single-driver rework: the DW_apb_i2c driver inverted from the classic blocking model to the XHAL asynchronous contract.

- **Control inversion**: the I-class starters program the peripheral and return; thread suspension and timeout ownership move to the shared synchronization layer. Every classic thread-context polling loop is gone — the busy wait becomes a single-shot gate, the abort becomes asynchronous, and the internal thread suspensions disappear.
- **Asynchronous abort**: initiated system-locked, completing on the transfer-abort interrupt, with the user-abort source silenced by construction through the terminal claim. A software abort-pending latch gates new transfers until the hardware confirms the flush, since the activity status reflects bus state rather than abort progress.
- **Terminal-event serialization** (reworked after review — see below): the claim, the driver-state transition and the waiter resume are one atomic unit, so the driver state *is* the arbitration point; only the user callback runs outside the lock.
- **Completion evidence**: decided from live FIFO state anchored on a stop-expected marker rather than the coalescing stop-detected flag, which could otherwise swallow the driver's own stop when cleared from a stale foreign-stop observation. Every flag clear is followed by re-evaluation in the same critical section.
- Classic interrupt engine, abort-cause decoding and timing derivation preserved; validation now rejects rates and counts the divider math cannot honor and enforces the documented spike-suppression relationships. A thread-context bus-clear helper is added for stuck-bus recovery (verifying both lines rise).

### Protocol invariants

1. **Claim atomicity** — generation parity, driver state and the waiter reference are only ever mutated together under one system-lock section.
2. **Single winner** — exactly one terminal path closes each generation (stop-completion, error, `stop_transfer`, `stop`); losers touch no state, buffers or waiter, only hardware flags.
3. **State/generation coherence** — an active driver state observed under the lock always carries an unclaimed generation, so `i2cStopTransferI` either legitimately wins or does nothing; it can never wake a waiter for an already-terminated transfer.
4. **Stop discipline** — completion never decided on the coalescing flag; evidence is monotonic for an open generation and re-evaluated after every clear.
5. **Buffer ownership** — the receive buffer is written only while a generation is open; residue is drained before the claim flips it, so after any claim the buffer belongs wholly to the caller.
6. **Start gating** — a start is accepted only with no abort pending, an idle bus and both enable handshakes confirmed; otherwise a busy return with no side effects.

## Review triage

- codex adversarial review round 1: **rejected as written** — 2 BLOCKER (claim/state non-atomicity opening a stop-vs-completion window; stop-detected clear-on-read race able to hang a transfer) and 3 MAJOR (activity status not an abort interlock; missing peripheral-disable handshake before the target-address write; bus clear false-succeeding with the clock line stuck low) plus 1 MINOR (spike-suppression derivation). All fixed by the rework above; the invariants were re-derived rather than patched.
- coderabbit CLI: rate-limited after today's volume — will be run and triaged before this leaves draft.

## Verification

- Build matrix clean: both chips × {default, checks+asserts+state-checker}, `USE_SMART_BUILD=no`; per-file stylecheck, OSAL gate, `git diff --check` clean; zero shared XHAL changes.
- On-hardware legs queued for the batched bench pass: NACK handling, mid-transfer abort, timeout-induced stop, and a bus scan against a bench device.

## Changelog

- XHAL: RP port gains the I2C driver with asynchronous aborts.

### DIFF
--- a/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/portab.c
+++ b/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/portab.c
@@ -82,6 +82,13 @@ const hal_adc_config_t portab_adc_config = {
   .dummy          = 0U
 };
 
+/**
+ * @brief   I2C0 configuration, standard mode at 100kHz.
+ */
+const hal_i2c_config_t portab_i2ccfg = {
+  .baudrate       = 100000U
+};
+
 /*===========================================================================*/
 /* Module local types.                                                       */
 /*===========================================================================*/
@@ -93,6 +100,20 @@ const hal_adc_config_t portab_adc_config = {
 /*===========================================================================*/
 /* Module local functions.                                                   */
 /*===========================================================================*/
+
+/*
+ * Assigns the I2C pads to the I2C function with the internal pull-ups
+ * enabled. The demo bus carries no external pull-ups, without them the
+ * lines would float and an unanswered address would be reported as a bus
+ * error instead of the expected acknowledge failure.
+ */
+static void portab_i2c_pads(void) {
+
+  palSetLineMode(PORTAB_LINE_I2C_SDA, PAL_MODE_ALTERNATE_I2C |
+                                      PAL_RP_PAD_PUE);
+  palSetLineMode(PORTAB_LINE_I2C_SCL, PAL_MODE_ALTERNATE_I2C |
+                                      PAL_RP_PAD_PUE);
+}
 
 /*===========================================================================*/
 /* Module exported functions.                                                */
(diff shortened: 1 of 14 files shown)


emolitor:
coderabbit CLI pass now complete on both commits of this branch: **zero findings**. That closes the review gate noted in the PR body — the codex round-one blockers were the only substantive findings on this stage, and the rework addressed them. On-hardware legs remain queued for the batched bench pass.

gdisirio (approved):

gdisirio:
There are few merge conflicts to be addressed on this one.

emolitor:
Rerunning the bench tests after resolving the merge conflicts.

PR #216, opened 2026-08-10
ARMv6-M: give PORT_IRQ_HANDLER bodies external linkage

### LINKED ISSUE #215: ARMv6-M: PORT_IRQ_HANDLER trampoline can fail to link under LTO partitioning
## Summary

`PORT_IRQ_HANDLER()` on ARMv6-M can produce an unresolved symbol at link time once an image grows large enough for LTO to split it across partitions. It is a size threshold, not a feature interaction, so a project can cross it by adding an unrelated driver.

Observed while building a dual-core RP2040 demo:

```
os/common/ports/ARMv6-M/smp/rp2/chcoresmp.c:311:(.text.Vector7C+0x8):
    undefined reference to `Vector7C_isr'
```

## Root cause

`os/common/ports/ARMv6-M/chcore.h` (~447-456) expands `PORT_IRQ_HANDLER(id)` into two objects: a `static __attribute__((used))` body named `id##_isr`, and a `naked` trampoline named `id` whose **inline assembly** reaches the body through a literal (`.word id##_isr`).

GCC's LTO partitioner does not see references that exist only inside inline asm. `used` keeps the body alive but does not tie it to the trampoline, so when the program is large enough to be split into ltrans partitions the body can be placed in — or privatized into — a different partition than the trampoline, and the literal has nothing to resolve against.

ARMv8-M is unaffected: that port does not use this trampoline shape.

## Reproduction and controls

Bisected with one fixed configuration (RP2040, `CH_CFG_SMP_MODE`, `CH_DBG_ENABLE_CHECKS`, `CH_DBG_ENABLE_ASSERTS`, LTO on) while growing the image by enabling additional drivers:

| Image | Result |
|---|---|
| baseline | links |
| + one driver group | links |
| + two driver groups | links |
| + three driver groups | **undefined reference** |

Two independent controls confirm the mechanism rather than the size:

- the identical failing configuration links cleanly with `USE_LTO=no`;
- the identical failing configuration links cleanly with LTO still enabled and `-flto-partition=none`.

## Suggested fix

Make the reference visible to the partitioner: give the body external linkage, add `__attribute__((externally_visible))` next to the existing `used`, or emit an explicit `.global` for the body symbol — whichever the maintainers prefer stylistically. Any of these prevents the partitioner from localizing or separating the body from the trampoline that names it.

Until then, any ARMv6-M SMP recipe with LTO enabled is one translation unit away from the same error, and the failure surfaces in whichever change happens to cross the threshold rather than in the change that introduced the pattern.

I have a fix prepared and verified against the reproducer plus the standard RP2040 targets, and can open a PR if that is useful — happy to leave the choice of mechanism to you.

emolitor:
Fixes #215.

## Summary

`PORT_IRQ_HANDLER(id)` expands to a `static` body named `id##_isr` plus a `naked` trampoline that reaches it through an inline-asm literal (`.word id##_isr`). GCC's LTO partitioner cannot see references that exist only inside inline asm, so once an image is large enough to be split across ltrans partitions the body can be privatised or placed in a different partition than the trampoline, and the link fails:

```
os/common/ports/ARMv6-M/smp/rp2/chcoresmp.c:311: undefined reference to `Vector7C_isr'
```

The body now has external linkage, routed through the existing `PORT_IRQ_HANDLER_LINKAGE` that the trampoline already uses, with `used` retained and a comment explaining why `static` must not come back.

## Why this mechanism

Two alternatives were tried and rejected:

- `__attribute__((externally_visible))` — GCC's `handle_externally_visible_attribute` warns *"attribute have effect only on public objects"* for a non-`TREE_PUBLIC` function, so on a `static` body it adds a `-Wattributes` warning **and** does not fix the linkage.
- an explicit `.global id##_isr` inside the trampoline's asm — that marks the symbol global in the *trampoline's* partition, where it is undefined; the definition in the other partition stays local and may be renamed.

External linkage is what a cross-partition reference actually requires: even if the partitioner knew about the reference, a local symbol cannot be bound across partitions.

Reusing `PORT_IRQ_HANDLER_LINKAGE` also closes a latent C++ problem — with the body `static`, a C++ translation unit would mangle its assembler name and the `.word` literal could never resolve.

## Scope

Exactly one port is affected. A repo-wide grep for `##_isr` matches only this file; every other port (ARMv6-M-ALT, ARMv7-M(-ALT), ARMv8-M-ML(-ALT/-TZ), ARMvx-M-SB, ARMv7-R, ARM, e200, AVR, RISCV-HAZARD3, the simulators and the templates) declares `PORT_IRQ_HANDLER(id)` as a plain function with no trampoline and no asm literal.

## Verification

- **Reproducer**: an RP2040 dual-core image with checks and assertions fails to link before the change and links after it, same tree, same command.
- **No code-generation change**: `nm -n` on `demos/RP/RT-RP2040-PICO` before/after differs only in the eight handler bodies moving from `t` to `T` **at identical addresses**; trampolines unchanged in address, type and binding; text size byte-identical at 99152. The only other diff is DWARF file-hash symbols in non-allocated sections.
- Builds clean, stock and with checks+assertions: `demos/RP/RT-RP2040-PICO`, `testhal/RP/RP2040/SIO-VALIDATION`; plus `testhal/RP/RP2350/SIO-VALIDATION` as an ARMv8-M sanity check that the shared header is undisturbed. No new warnings.

Left as a draft so the mechanism can be swapped if you prefer a different one — the reproducer and the `nm` control make it easy to re-verify any variant.

## Changelog

- ARMv6-M: interrupt handler bodies are given external linkage so LTO partitioning cannot separate them from their trampolines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,12 @@ applied to a maintenance branch are marked *(backported to 21.11.6)*.
 
 ### Fixed
 
+- ARMv6-M could fail to link with `-flto` (an "undefined reference" to the
+  interrupt handler body) once an image grew large enough for the LTO
+  partitioner to split it; the `PORT_IRQ_HANDLER` body now has external
+  linkage and a reserved name so the trampoline's assembler reference
+  resolves across partitions
+  ([#216](https://github.com/chibios-upstream/chibios/pull/216)).
 - Serial over USB input remained inactive after a USB suspend/wakeup cycle on
   STM32 OTG devices. The generic USB driver terminates all pending
   transactions on suspend and the OTG LLD also disables the endpoints in
--- a/os/common/ports/ARMv6-M/chcore.h
+++ b/os/common/ports/ARMv6-M/chcore.h
@@ -444,12 +444,20 @@ struct port_context {
  *          body GCC is free to place the body and the trampoline in different
  *          @p ltrans partitions, or to privatise the local symbol under a new
  *          name; the literal then has nothing to bind to and the link fails
- *          with @p "undefined @p reference @p to @p Vector7C_isr" as soon as
- *          an image grows large enough for @p -flto to split it.
+ *          with @p "undefined @p reference" as soon as an image grows large
+ *          enough for @p -flto to split it.
  *          @p __attribute__((used)) only pins the body against dead-code
  *          elimination, it does not make the asm-only reference visible: do
  *          not re-add @p static here. @p PORT_IRQ_HANDLER_LINKAGE keeps the
  *          assembler name unmangled when a handler is compiled as C++.
+ *          @n@n
+ *          Because that linkage publishes one global symbol per handler, the
+ *          body is deliberately named @p __port_irq_body_<id> rather than
+ *          @p <id>_isr, keeping it in the implementation-reserved namespace
+ *          already used by @p __port_irq_epilogue() and the other port
+ *          internals so it cannot collide with an application symbol. The
+ *          name is part of the macro contract: the forward declaration, the
+ *          definition and the @p ".word" literal must be kept in step.
  */
 #if defined(__GNUC__) || defined(__DOXYGEN__)
   #ifdef __cplusplus
@@ -459,17 +467,17 @@ struct port_context {
   #endif
   #define PORT_IRQ_HANDLER(id)                                              \
     PORT_IRQ_HANDLER_LINKAGE __attribute__((used))                         \
-    void id##_isr(uint32_t _saved_lr);                                     \
+    void __port_irq_body_##id(uint32_t _saved_lr);                         \
     PORT_IRQ_HANDLER_LINKAGE __attribute__((naked, used))                  \
     void id(void) {                                                        \
       __asm volatile ("mov   r0, lr            \n\t"                       \
                       "ldr   r1, 1f            \n\t"                       \
                       "bx    r1                \n\t"                       \
                       ".balign 4               \n\t"                       \
-                      "1: .word " #id "_isr    \n\t");                     \
+                      "1: .word __port_irq_body_" #id "\n\t");             \
     }                                                                      \
     PORT_IRQ_HANDLER_LINKAGE __attribute__((used))                         \
-    void id##_isr(uint32_t _saved_lr)
+    void __port_irq_body_##id(uint32_t _saved_lr)
 #else /* IAR (__ICCARM__) / ARMCC5 (__CC_ARM): EXC_RETURN captured in the
          prologue via a compiler intrinsic; no trampoline.
          TODO: verify whether these toolchains' duplicate-function merging
--- a/os/common/ports/ARMv6-M/chcore.h
+++ b/os/common/ports/ARMv6-M/chcore.h
@@ -436,6 +436,20 @@ struct port_context {
  *          of the only unconditional Thumb branch available on ARMv6-M. The
  *          address literal is emitted immediately after the trampoline so the
  *          Thumb-1 literal load is independent of assembler pool placement.
+ *          @n@n
+ *          The body must have external linkage. The trampoline reaches it
+ *          only through the assembler name in the @p ".word" literal, and a
+ *          reference that lives inside inline assembly is invisible to the
+ *          compiler and, critically, to the LTO partitioner. With a @p static
+ *          body GCC is free to place the body and the trampoline in different
+ *          @p ltrans partitions, or to privatise the local symbol under a new
+ *          name; the literal then has nothing to bind to and the link fails
+ *          with @p "undefined @p reference @p to @p Vector7C_isr" as soon as
+ *          an image grows large enough for @p -flto to split it.
+ *          @p __attribute__((used)) only pins the body against dead-code
+ *          elimination, it does not make the asm-only reference visible: do
+ *          not re-add @p static here. @p PORT_IRQ_HANDLER_LINKAGE keeps the
+ *          assembler name unmangled when a handler is compiled as C++.
  */
 #if defined(__GNUC__) || defined(__DOXYGEN__)
   #ifdef __cplusplus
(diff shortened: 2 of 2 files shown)


emolitor:
Ready for review now.

emolitor:
@gdisirio to review as this is common ARMv6 startup code.

emolitor:
Pushed one additional commit (no force-push, since this is out of draft): the handler bodies move from `<id>_isr` to the implementation-reserved `__port_irq_body_<id>`. Making them global was necessary, but it put one ordinary-namespace symbol per handler into every image, which an application could legitimately already define. The prefix follows the port's own convention (`__port_irq_epilogue`, `__port_switch`, `__port_thread_start`, ...).

I also replaced the weak evidence in the original description. The earlier control — `nm` on one image at one optimisation level — proved much less than it appeared to: it inspected neither relocations, nor section GC, nor visibility, nor alternate IPA decisions, and it did not deterministically cover the failure.

**Deterministic reproduction.** `-flto-partition=max` reproduces the undefined reference in a *small* image, so no size threshold is needed. 16 legs — two projects (`demos/RP/RT-RP2040-PICO`, `testhal/RP/RP2040/SIO-VALIDATION`) × `-Os`/`-O2` × `--gc-sections` on/off × pre-fix/fixed:

- pre-fix: **fails in all 8**, reproducing the original text verbatim including ``chcoresmp.c:311: undefined reference to `Vector7C_isr'``
- fixed: **links in all 8**

**Handlers did not become GC roots.** A `CH_IRQ_HANDLER` with no vector-table slot is still dropped under `--gc-sections`, before and after — with a negative control (`USE_LINK_GC=no` keeps it) proving the check is not vacuous. Confirmed separately that `used` does not set `SHF_GNU_RETAIN` on GCC 15.2.1, so the linker remains free to drop the sections.

**C++ expansion.** A C++ translation unit instantiating the macro emits `T __port_irq_body_VectorA8` — global and unmangled — while a mangled static in the same unit confirms it really went through the C++ front end. Pre-fix, that same unit fails to link even *without* partition stress, because the static body was mangled while the assembly literal was not. That is the latent C++ defect this also closes.

**No code-generation change.** Same tree, same path, pinned `SOURCE_DATE_EPOCH`: `ch.bin` byte-identical, `objdump -dr` identical once the rename is canonicalised, allocated sections identical, and the symbol-table delta is exactly the eight bodies moving `t` → `T` at unchanged addresses. The only other movement is in non-allocated debug sections.

ARMv8-M sanity leg contains zero `__port_irq_body_` symbols, confirming the macro is ARMv6-M-only and the shared header is undisturbed elsewhere.

emolitor:
A design question this fix raises. This defect is specific to the classic ARMv6-M trampoline. `ARMv6-M-ALT` expands `PORT_IRQ_HANDLER(id)` to a plain `void id(void)` with a common dispatcher — no naked trampoline, no `id##_isr` body, no assembly literal — so the whole failure mode is absent there. RP2350 is already on `ARMv8-M-ML-ALT`; RP2040 is the odd one out on the classic port.

Measuring the two ARMv6-M variants on the same kernel (a build with two threads, a semaphore and a varying number of `CH_IRQ_HANDLER`s), the marginal ROM cost per handler is **16 bytes on ALT against 96 bytes on classic**, break-even at roughly three handlers; at sixteen handlers ALT is about 1 KB smaller. Working-area sizes are byte-identical between the two, so on ARMv6-M this is a code-size and structural difference rather than a RAM one.

What stops RP2040 moving today is that `ARMv6-M-ALT` has no `smp/rp2` layer and no `port_rp2.mk`, and the RP2 SMP code leans on port internals that only exist in the classic port — `__port_smp_init`, `__port_spinlock_take`/`_release`, and `__port_flash_lockout`/`__port_lockout_other_ready` (the last of which our XIP flash safety depends on). `__port_get_irq_status` and `__port_irq_enabled` are the only ones already present. So it is a port-layer job, not a recipe switch, and it sits squarely in your ALT migration rather than anywhere I would go unilaterally. What's odd is I thought we had done this work earlier but not seeing it in the tree. I will excavate my old SVN tree to see if I missed something.

This is not a reason to hold #216: the classic port stays in the tree, and any sufficiently large image built with LTO hits the same trampoline problem whether or not it is SMP or RP.

chibios-sheriff (commented):

Scope: give the ARMv6-M `PORT_IRQ_HANDLER` body external linkage (and a reserved-namespace name) so the trampoline's `.word` literal resolves under `-flto`. Base `master`.

> not the full review yet.

Correct, and the doxygen rationale is excellent. The trampoline reaches the handler body only through the assembler name in the `.word` literal, which is invisible to the compiler and the LTO partitioner; a `static` body can be placed in a different `ltrans` partition (or privatised under a new local name), leaving the literal nothing to bind to — hence the "undefined reference" once `-flto` splits a large image. Giving the body external linkage fixes that, and `__attribute__((used))` alone can't (it pins against DCE but doesn't publish the symbol).

The details check out:
- The forward declaration, the definition, and the `.word` literal all use `__port_irq_body_##id` consistently.
- `PORT_IRQ_HANDLER_LINKAGE` (`extern "C"` under C++) keeps the name unmangled so it matches the asm literal.
- The `__port_irq_body_<id>` name sits in the `__port_*` implementation-reserved namespace (alongside `__port_irq_epilogue`), so publishing one global per handler won't collide with application symbols.
- No API/ABI impact: the vector function `id()` is unchanged; the body was `static` (no external references existed), so making it external + renaming breaks nothing for callers.

Mechanical gates: `stylecheck.py` clean; `testhal/STM32/multi/USB_RAW` on `stm32g0b1re_nucleo64` (Cortex-M0+) links cleanly with `USE_LTO=yes`, ARM GCC 14.3.1.

Reviewer-owned changelog: this fixes a real ARMv6-M link failure under LTO, so it takes a `readme.txt` entry — I'll push it to this branch.

gdisirio (approved):

PR #217, opened 2026-08-10
XHAL: restore the RISC-V leg of the RP multi-target demo

emolitor:
## Summary

`demos/RP/RT-XHAL-RP-MULTI/make/rp2350_pico2_riscv.make` has not compiled since #210 merged — 17 errors (`adcsample_t`, `pwmcnt_t`, `PWM_EVENT_PERIOD`, `PORTAB_PWM`, `PORTAB_ADC`, `pwmEnableEvents`, ...). Because the demo's top-level Makefile globs `make/*.make`, a bare `make` in that directory fails.

This is merge-order debt rather than a code defect: #208 extended the **shared** `main.c` with the PWM and ADC exercises, while #210 — authored against pre-#208 master and merged after it — brought a RISC-V config tree with `HAL_USE_PWM`/`HAL_USE_ADC` false, no `PORTAB_PWM`/`PORTAB_ADC` symbols, and a `platform_riscv.mk` without those drivers. #210's own PR body flagged the reconciliation as owed at rebase time; the merge landed before it happened.

Confirmed against pristine master with master's own `main.c`, so it is not an artefact of any pending branch.

## Change

- `os/xhal/ports/RP/RP2350/platform_riscv.mk`: add the ADCv1 and PWMv1 driver recipes.
- `cfg/rp2350_pico2_riscv/`: enable ADC and PWM, add the matching knobs, the `PORTAB_PWM`/`PORTAB_ADC` symbols and configurations, and move GP25 to the PWM alternate function (dropping the now-wrong LED write).

The RISC-V recipe deliberately keeps the classic port's exclusions rather than copying the ARM list wholesale:

| Driver | In RISC-V recipe? | Reason |
|---|---|---|
| ADCv1, PWMv1 | yes | needed by the shared demo; register-level LLDs whose only architecture contact is the Hazard3 NVIC shim, and whose default IRQ priority 3 is legal on Xh3irq (levels 0..3) |
| TIMERv1 | no | its recipe adds a second `hal_st_lld.c` on top of the Hazard3 one; the Hazard3 port owns its core-local timer |
| EFLv1 + flash safety | no | `rp_flash_safety.c` needs `__port_flash_lockout()`, which exists only in the ARM SMP ports; the Hazard3 port has no lockout services |
| RTCv2, USBv1 | no | nothing in-tree exercises them on RISC-V, and enabling RTCv2 needs an IRQ-priority knob this config tree does not define — deliberate future steps, not part of unbreaking the build |

After the change the ARM and RISC-V Pico 2 config trees differ only in `xmcuconf.h`, by the five ARM-only SysTick/TIMER0 knobs.

## Verification

- The failure was reproduced first, then fixed.
- **A bare `make` in `demos/RP/RT-XHAL-RP-MULTI` now exits 0 across all three recipes** — that is the point of the change.
- All three legs also build with checks and assertions enabled; `testxhal/SPI` RISC-V builds as a `platform_riscv.mk` regression.
- Remaining warnings are pre-existing and were confirmed on stock master (the two ARM RWX-segment linker notes, and one `-Wmaybe-uninitialized` in the RT test suite on the RISC-V leg).
- Style, `git diff --check` and the OSAL lexical gate all clean.

## Follow-up this implies

The same config tree will need the same treatment when I2C (#213) and the pending RTC demo exercise start being exercised from the shared `main.c`. RTCv2 on RISC-V in particular wants a deliberate decision — the classic RISC-V recipe omits it — so I have left that to the branch that needs it rather than inheriting my omission silently.

## Changelog

- XHAL: restored the RISC-V leg of the RP multi-target demo after the PWM and ADC exercises landed.

### DIFF
--- a/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2350_pico2_riscv/portab.c
+++ b/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2350_pico2_riscv/portab.c
@@ -34,6 +34,54 @@
 /* Module exported variables.                                                */
 /*===========================================================================*/
 
+/**
+ * @brief   PWM configuration for the LED slice.
+ * @details The slice counts at 1MHz over a 1000 ticks period giving a
+ *          1kHz PWM frequency, the LED on GP25 is driven by the B
+ *          channel. Events are enabled at runtime by the application.
+ */
+const hal_pwm_config_t portab_pwm_config = {
+  .frequency      = 1000000U,
+  .period         = 1000U,
+  .enabled_events = 0U,
+  .channels       = {
+    {
+      .mode       = PWM_OUTPUT_DISABLED
+    },
+    {
+      .mode       = PWM_OUTPUT_ACTIVE_HIGH
+    }
+  },
+  .dummy          = 0U
+};
+
+/**
+ * @brief   Conversion groups of the portability ADC configuration.
+ * @details A single conversion group sampling the on-die temperature
+ *          sensor at the free-running rate, the sensor bias is enabled
+ *          by the driver for the duration of the conversion.
+ */
+static const adc_conversion_groups_t portab_adc_groups = {
+  .grpsnum        = 1U,
+  .grps           = {
+    {
+      .num_channels = 1U,
+      .channel      = ADC_CHANNEL_TEMPSENSOR,
+      .rrobin       = 0U,
+      .div          = 0U,
+      .ts_enabled   = true
+    }
+  }
+};
+
+/**
+ * @brief   ADC configuration carrying the temperature sensor group.
+ */
+const hal_adc_config_t portab_adc_config = {
+  .grps           = &portab_adc_groups,
+  .dummy          = 0U
+};
+
 /*===========================================================================*/
 /* Module local types.                                                       */
 /*===========================================================================*/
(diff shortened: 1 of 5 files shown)


emolitor:
Review pass done: coderabbit zero findings; codex found no blocker and no architecture defect, and confirmed the interrupt numbering, the absence of any ARM-only intrinsic or system-timer dependency in ADCv1/PWMv1/DMAv1, and that `RP_ADC_ADC1_DMA_PRIORITY` is a DMA arbitration priority rather than an interrupt level. Two things came out of it that are worth recording.

**Fixed — a priority value that does not mean the same thing on both cores.** I had carried `RP_ADC_ADC1_DMA_IRQ_PRIORITY 3` across from the ARM configuration. The ARM core has sixteen priority levels, so 3 sits near the top; Hazard3 has four, so the same literal makes it the **least urgent interrupt in the system**. The conversions run free into an eight-entry FIFO and stop only when that interrupt is served, so the copied value quietly created an overflow risk under interrupt load. It is now 1 on the RISC-V configuration, with the reason recorded next to it. PWM at 1 kHz stays at 3, where lowest urgency is harmless.

**Corrected wording.** My description said the excluded drivers were incompatible with Hazard3. That is too strong and only accurate for the built-in EFL lockout, which genuinely needs `__port_flash_lockout()` services that only the ARM SMP ports provide — EFL itself supports application-provided safety hooks, and RTCv2 and USBv1 look architecture-compatible (RTC merely wants its priority knob, USB already has explicit RISC-V handling). They are **unwired and untested here, not incompatible**, and enabling any of them deserves its own change with a target that exercises it.

**On what this PR does and does not claim.** ADCv1 and PWMv1 can be claimed to build and to be statically compatible with Hazard3. They cannot yet be claimed to *work* on Hazard3 silicon — no RISC-V hardware run has happened, and a green build (or even green kernel test suites) does not exercise the fade, the wrap counter or the ADC DMA path. That evidence is queued in a bench batch covering the recently merged stages, and I will post it here. If you would rather not carry an unvalidated claim in the interim, the honest framing is that this restores the build; the runtime claim comes with the bench results.

Bare `make` in the demo directory still exits 0 across all three recipes after the priority change.

gdisirio (approved):

PR #218, opened 2026-08-10
XHAL: exercise the RP RTC drivers from the multi-target demo

emolitor:
## Summary

The RTC drivers landed in #211 without any on-target exercise: the demo self-check written alongside them was still an uncommitted working-tree change when the PR merged, so master has `HAL_USE_RTC FALSE` and the merged drivers have **no runtime coverage at all**. This restores it.

Adversarial review of #211 had also flagged the gap explicitly — the existing `testxhal/RTC` builds STM32 register images for its alarms and cannot exercise the RP contract, so the demo is the only vehicle available today.

The self-check runs once after the test suites and prints one line per step, so a bench capture can be read directly:

1. read the time *before* setting it, and confirm it reports the not-set condition (documented behaviour on this port, verified rather than treated as a failure);
2. set a known timestamp and read it back, confirming the round trip;
3. arm a one-shot alarm a few seconds out, wait for the callback, and report the measured delay against the RTC's own clock;
4. read the alarm back afterwards and confirm it reads as disarmed;
5. print a `rtc: PASS` / `rtc: FAIL <what>` summary.

## Notes for review

- **Ordering is a real constraint, not a style choice.** The check measures alarm latency and asserts it lands close to the expected lead, so it runs *before* the PWM and ADC exercises start — the 1 kHz PWM wrap interrupt would otherwise perturb the very timing it measures.
- **Per-chip divergence is deliberate.** The two config trees differ in the retained-time flag and the alarm timeout; everywhere else in this demo they are byte-identical, so this is the one place the usual "copy one over the other" shortcut is wrong.
- Master's console helpers are reused as-is; epoch and alarm values sit close to `INT32_MAX`, so they print through an unsigned helper rather than the existing signed one.

## Verification

Both targets build clean with default flags and with checks and assertions enabled, plus `USE_SMART_BUILD=no`; style, `git diff --check` and the OSAL gate are clean.

On-hardware results will be posted here from the bench batch that is working through the recently merged stages — this branch is the piece that makes the RTC part of that batch possible.

## Changelog

- XHAL: the RP multi-target demo exercises the RTC drivers.

### DIFF
--- a/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/portab.h
+++ b/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/portab.h
@@ -49,6 +49,24 @@
 #define PORTAB_I2C                  I2CD0
 #define PORTAB_LINE_I2C_SDA         4U
 #define PORTAB_LINE_I2C_SCL         5U
+#define PORTAB_RTC                  RTCD1
+
+/*
+ * The RP2040 RTC block sits behind the RESETS block and is held in reset
+ * at every chip reset, its counters always come up cleared. A date/time
+ * read taken before a date/time has been set therefore always reports the
+ * not-set condition.
+ */
+#define PORTAB_RTC_TIME_RETAINED    FALSE
+
+/*
+ * Alarm lead time, in RTC seconds, and the wall clock budget allowed for
+ * it to elapse. The RTC 1Hz reference is divided down from clk_rtc, which
+ * is crystal derived through the USB PLL, so RTC seconds and wall clock
+ * seconds match and a small margin is enough.
+ */
+#define PORTAB_RTC_ALARM_LEAD_S     3U
+#define PORTAB_RTC_ALARM_TIMEOUT_MS 6000U
 
 /*===========================================================================*/
 /* Module pre-compile time settings.                                         */
--- a/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/xhalconf.h
+++ b/demos/RP/RT-XHAL-RP-MULTI/cfg/rp2040_pico/xhalconf.h
@@ -56,7 +56,7 @@
 #define HAL_USE_ICU                         FALSE
 #define HAL_USE_MMC_SPI                     FALSE
 #define HAL_USE_PWM                         TRUE
-#define HAL_USE_RTC                         FALSE
+#define HAL_USE_RTC                         TRUE
 #define HAL_USE_SDC                         FALSE
 #define HAL_USE_SIO                         TRUE
 #define HAL_USE_SPI                         FALSE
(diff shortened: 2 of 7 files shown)


emolitor:
Rebased onto current master and pushed two follow-ups on top of the autofix commit (kept, not squashed).

**Alarm left armed on the timeout path.** The timeout branch dropped the callback but left the alarm armed in hardware, so a late one-shot could still fire after the self-check had already reported failure. It now disarms first and drops the callback second, so the source is silenced before the handler goes away.

**RISC-V leg no longer built.** This is the same merge-order debt the RISC-V target already hit once with the PWM and ADC additions: the RTC self-check went into the shared `main.c` unconditionally, but `RP2350/platform_riscv.mk` does not build RTCv2 and the RISC-V configuration has `HAL_USE_RTC` `FALSE`. The RTC state, the time-printing helpers used only by it, and the self-check are now behind `HAL_USE_RTC`. Guarding is the right call rather than enabling the driver there — the RISC-V platform intentionally drops RTCv2.

The remaining coderabbit note about an uninitialised `rtc_alarm_t` does not hold up: `rtc_alarm_t` has exactly one member, `alrmr`, and it is assigned immediately before the `rtcSetAlarm()` call, so nothing indeterminate is ever read. No change made.

**Verification.** All three targets rebuilt clean, no warnings:

| Target | text | bss |
|---|---|---|
| `rp2040_pico` | 109860 | 268736 |
| `rp2350_pico2` | 112352 | 531408 |
| `rp2350_pico2_riscv` | 91822 | 530176 |

The two ARM images are identical in size before and after the guards, confirming they are inert where RTC is enabled. Per-file `stylecheck.py` and `git diff --check` clean. The behavioural change is confined to a failure path, so the RP2350 bench evidence already posted for the alarm arm/fire path still stands.

emolitor:
Rebased onto master after #213 merged. The conflict was substantive rather than textual — both PRs extend the same demo `main.c` — so a note on how it was resolved:

- **Constants, helpers and functions**: additive on both sides, both kept.
- **`print_udec()` doc comment**: master's wording is the general one and was kept, with a line added noting the RTC epoch seconds are one such value.
- **`print_udec()` guard**: my earlier RISC-V guard wrapped this helper, which was correct when only the RTC used it. Master's `i2c_scan()` now calls it five times, so the guard was moved to start at `print_dec2()` and `print_udec()`/`print_hex2()` are left unguarded. Git had reapplied my guard hunks by context matching and, with the new I2C code in between, they had landed around `print_udec()`, `print_hex2()`, `i2c_recover()` and `i2c_scan()` — all of which are I2C, not RTC. Corrected so each `HAL_USE_RTC` region covers only RTC code.

**This PR is stacked on #220.** The RISC-V leg of the demo does not build on master right now — #213 added the I2C scan unguarded and `platform_riscv.mk` does not build `I2Cv1`. That is a pre-existing break, not something this branch introduces, and #220 fixes it. With #220 applied locally on top of this branch, all three targets build clean:

| Target | text |
|---|---|
| `rp2040_pico` | 113724 |
| `rp2350_pico2` | 115768 |
| `rp2350_pico2_riscv` | 91822 |

One thing to watch at merge time: #220 guards `print_udec()` behind `HAL_USE_I2C`, and this branch's RTC printers call it. With both drivers on (the ARM targets) or both off (RISC-V) that is fine, which is why the stack builds, but the guard should be widened to cover either driver when the two land, so an RTC-on/I2C-off configuration cannot break. I will do that on whichever rebases second.

emolitor:
The `print_udec()` guard interaction flagged above is now handled on #220's side, so nothing is left pending here for it: `print_udec()` is guarded on either driver there and this branch's RTC printers can call it whatever `HAL_USE_I2C` is set to. Merge order between the two no longer matters.

gdisirio (approved):

PR #220, opened 2026-08-10
XHAL: keep the RP demo RISC-V leg building without the I2C driver

emolitor:
**Master is currently broken for `demos/RP/RT-XHAL-RP-MULTI` on RISC-V.** At `b63efdf6d4` (#213 merge):

```
main.c:181:12: error: 'PORTAB_I2C' undeclared (first use in this function)
main.c:182:9:  error: implicit declaration of function 'portab_i2c_bus_clear'
main.c:184:33: error: 'portab_i2ccfg' undeclared (first use in this function)
main.c:200:3:  error: unknown type name 'i2cflags_t'
```

The bus scan went into the shared `main.c` unconditionally, but `RP2350/platform_riscv.mk` does not build `I2Cv1` and the RISC-V configuration sets `HAL_USE_I2C` `FALSE`.

## Fix

Guards behind `HAL_USE_I2C`: the scan constants, the `found[]` array, `print_udec()`/`print_hex2()` (used only by the scan), `i2c_recover()`/`i2c_scan()`, and in `main()` the `iteration` counter, the driver activation and the periodic call.

Guarding rather than enabling the driver on RISC-V is the conservative choice here — it restores the previous state exactly. Enabling I2C for the RISC-V platform is defensible on the hardware (DW_apb_i2c is not core-specific) but it is a port change that deserves its own build and bench evidence, so it is noted below rather than bundled in.

## Verification

| Target | Result |
|---|---|
| `rp2040_pico` | builds, 108640 text, no `main.c` warnings |
| `rp2350_pico2` | builds, 111376 text, no `main.c` warnings |
| `rp2350_pico2_riscv` | **builds again**, 91822 text, no `main.c` warnings |

The guards are provably inert where I2C is enabled: the `rp2350_pico2` binary is **byte-identical** with and without them apart from the two bytes of compile timestamp in the banner string. The RISC-V image at 91822 text matches the size of the last RISC-V build taken before the I2C work landed.

Per-file `stylecheck.py` and `git diff --check` clean.

## Worth a decision

This is the third time an addition to this demo has broken the RISC-V leg — after #208/#210 (fixed by #217, PWM and ADC) and the RTC work in #218. One `main.c` serves three targets and the RISC-V platform deliberately carries a smaller driver set (no `TIMERv1`, `RTCv2`, `EFLv1`, `USBv1`, `I2Cv1`), so every driver-touching addition needs a guard and CI does not cover these targets. Two options worth considering: add the RP demo targets to the CI sweep so this fails in the PR rather than on master, and/or close the gap by adding `I2Cv1` to `platform_riscv.mk` where the hardware allows it.

### DIFF
--- a/demos/RP/RT-XHAL-RP-MULTI/main.c
+++ b/demos/RP/RT-XHAL-RP-MULTI/main.c
@@ -124,11 +124,13 @@ static void print_dec(BaseSequentialStream *stream, int32_t value) {
   stmWrite(stream, (const uint8_t *)&buf[i], sizeof buf - i);
 }
 
-#if (HAL_USE_I2C == TRUE) || defined(__DOXYGEN__)
+#if (HAL_USE_I2C == TRUE) || (HAL_USE_RTC == TRUE) || defined(__DOXYGEN__)
 /*
  * Writes an unsigned decimal number on the stream. Counters and any
  * value that can exceed the signed range are printed through this
  * helper, passing them to the signed printer would misrepresent them.
+ * Both the bus scan and the RTC check need it, so it is available to
+ * either of them.
  */
 static void print_udec(BaseSequentialStream *stream, uint32_t value) {
   char buf[10];
@@ -142,7 +144,9 @@ static void print_udec(BaseSequentialStream *stream, uint32_t value) {
   } while (value > 0U);
   stmWrite(stream, (const uint8_t *)&buf[i], sizeof buf - i);
 }
+#endif /* HAL_USE_I2C == TRUE || HAL_USE_RTC == TRUE */
 
+#if (HAL_USE_I2C == TRUE) || defined(__DOXYGEN__)
 /*
  * Writes the low byte of a value as a two digits hexadecimal number on
  * the stream.
--- a/demos/RP/RT-XHAL-RP-MULTI/main.c
+++ b/demos/RP/RT-XHAL-RP-MULTI/main.c
@@ -23,6 +23,7 @@
 
 #include "portab.h"
 
+#if (HAL_USE_I2C == TRUE) || defined(__DOXYGEN__)
 /*
  * Scanned 7-bit address range, the reserved addresses outside it are not
  * probed.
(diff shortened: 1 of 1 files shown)


emolitor:
Added one commit (no force-push, this is out of draft) closing a merge-order trap between this PR and #218.

`print_udec()` was guarded together with `print_hex2()` behind `HAL_USE_I2C`, because the bus scan is its only caller on master. #218's RTC self-check calls the same helper, so a configuration with the RTC enabled and I2C disabled would have lost it — the two PRs build fine together only because every current configuration has both drivers on or both off. Rather than leave that to whichever branch rebases second, `print_udec()` is now guarded on either driver, while `print_hex2()` stays behind `HAL_USE_I2C` alone since only the scan uses it.

**Nothing built today changes.** Every demo configuration on master currently has `HAL_USE_RTC` `FALSE`, so the widened condition is still decided by `HAL_USE_I2C` exactly as before, and that is visible in the artefacts: all three targets rebuild to the same sizes as the previous commit (108640 / 111376 / 91822 text, no `main.c` warnings), and the `rp2350_pico2` binary differs from the earlier build only in the five bytes of compile timestamp inside the banner string.

emolitor:
Review gates now run on this branch, both clean, with one caveat worth stating rather than burying.

**coderabbit CLI**: no findings.

**codex adversarial review**: guards verified complete across all four combinations of `HAL_USE_I2C` × `HAL_USE_RTC` — no I2C-dependent identifier left unguarded, none referenced from unguarded code, and the `iteration` local's declaration, initialisation and use all carry the same condition.

It did flag one real consequence of the previous commit. In an `HAL_USE_I2C` `FALSE` / `HAL_USE_RTC` `TRUE` configuration, `print_udec()` is compiled with no caller in *this branch alone* — the RTC caller arrives with #218 — so that combination would emit `-Wunused-function`. No configuration in the tree is affected: every demo configuration on master has `HAL_USE_RTC` `FALSE`, and the three real targets build warning-free.

I have left it that way deliberately. The alternative is to guard `print_udec()` on `HAL_USE_I2C` alone, which is warning-clean in isolation but means a broken build if #218 lands first and nobody widens the guard by hand. Given both PRs are ready and either merge order is possible, a warning in a configuration nobody builds seemed the better exposure than a build that depends on merge order. Happy to flip it if you would rather each PR be self-contained and handle the widening at merge time.

gdisirio (approved):

PR #222, opened 2026-08-11
RP PIOv1: add block level interrupt, callback and GPIO routing APIs

begeistert:
## Summary

`IRQ0_INTE`/`IRQ1_INTE`, the block interrupt line and IO_BANK0 FUNCSEL are PIO block resources, but the driver only reached them through a state machine handle. A consumer that routes its pins or enables a block interrupt before allocating any state machine — or that needs work done exactly once per interrupt — had no function to call. The nanoFramework PIO binding (nanoframework/nf-interpreter#3449) open-codes all three today: raw INTE writes, raw once-per-interrupt acknowledgement in its callback, and raw IO_BANK0/PADS_BANK0 writes including a hand-rolled RP2350 isolation sequence.

This PR:

- adds `pioEnableInterruptX()`/`pioDisableInterruptX()` taking the block descriptor and the same `PIO_IRQ_*` masks as the per state machine pair: enable programs the current core's INTE (selected by `SIO->CPUID`), disable clears both cores' unconditionally, matching the existing Sm semantics;
- adds `pioSetBlockCallback()`/`pioSetBlockCallbackI()`: a callback invoked exactly once per interrupt of the block, before the per state machine callbacks, with the same `IRQn_INTS` word. The driver acknowledges nothing itself and `IRQn_INTS` is derived from INTR and the FIFO levels, so unless a handler clears the source the interrupt re-fires; with only per state machine callbacks that acknowledgement is either duplicated N ways or coordinated out of band. A block with no callback registered behaves exactly as before (NULL fast-path in the ISR);
- adds block level GPIO routing: `pioSetPinFunctionX()`, `pioGpioRoutePadX()` and `pioGpioRouteX()` take the block descriptor. The existing `pioSmSetPinFunctionX()`, `pioGpioInitPadX()` and `pioGpioInitX()` keep their names, signatures and emitted register writes and now forward to the block versions — their bodies only ever consumed `smp->block`, so this is a pure refactor for existing callers. The RP2350 glitch-free pad isolation sequence now lives in one place, `pioGpioRoutePadX()`.

PIO-VALIDATION grows tests 15..17: block interrupt enables with no state machine allocated (current-core INTE only, both-core disable, INTS follows the forced flag), block callback ordering and NULL removal, and routing before allocation with pad pull readback plus the RP2350 ISO drop. Test numbers 15..22 are reserved across this PIO API series; the header note in `main.c` documents the reservation, this branch carries only its own tests.

## Related

- Issue(s): —
- Notes for reviewers: the block routing pair could not reuse the `pioGpioInitPadX`/`pioGpioInitX` names (taken by the state machine versions, no overloading in C); `pioGpioRoute*X` follows the "Routes a GPIO pin" wording the existing doc already uses. If you would rather rename the Sm variants and free the short names for the block, happy to rework. Independent of the other three PRs in this series (runtime setters/readback, patching/synchronized enable, instruction encoders), which will follow.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

- [x] Style check passed on changed `.c`/`.h` files (`tools/style/stylecheck.py` on `rp_pio.c`, `rp_pio.h` and both `main.c`)

## Testing

- RP2350 leg run on a Pico 2: **139 pass, 0 fail**; RP2040 leg run on a Pico W: **123 pass, 0 fail** (reports captured from UART0).

## Compatibility and Risk

The three Sm-handle GPIO functions are unchanged in name, signature and behavior (one-line forwarders). The block callback adds one NULL check to the ISR fast path. Test 15 includes a sub-case asserting that a flag raised through `IRQ_FORCE` is visible in `pioIrqGetX()` and clearable with `pioIrqClearX()`, as their documentation states; if the device run contradicts that, a `pioIrqGetForcedX()` accessor will be added to this PR.

### DIFF
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
@@ -504,10 +504,14 @@ void pioSmFreeI(const rp_pio_sm_t *smp) {
 
   /* Reset PIO block only if no state machines remain allocated and no
      programs are loaded, resetting while instruction memory is allocated
-     would wipe loaded programs behind the bookkeeping's back.*/
+     would wipe loaded programs behind the bookkeeping's back. The reset
+     wipes the INTE routing, so the block callback is dropped with it:
+     keeping the pointer past the point it can ever fire again would
+     leave a stale registration behind.*/
   if (((pio.blocks[b].c0_allocated_mask |
         pio.blocks[b].c1_allocated_mask) == 0U) &&
       (pio.blocks[b].imem_allocated == 0U)) {
+    pio.blocks[b].block.func = NULL;
     rp_peripheral_reset(smp->block->resets_mask);
   }
 }
@@ -735,6 +739,11 @@ void pioSmInit(const rp_pio_sm_t *smp, uint32_t initial_pc,
  *          natural place to do it, being the only handler guaranteed to
  *          run exactly once per interrupt.
  * @note    Passing @p NULL removes the callback.
+ * @note    The callback can only be invoked while the current core has
+ *          at least one state machine allocated, which is what keeps the
+ *          PIO interrupt vector enabled. It is dropped automatically
+ *          when the block becomes fully idle and is reset, together with
+ *          the INTE routing.
  *
  * @param[in] block     pointer to the PIO block descriptor
  * @param[in] func      callback function, can be @p NULL
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
@@ -94,6 +94,19 @@ static struct {
      */
     uint32_t        imem_allocated;
     /**
+     * @brief   Block level IRQ redirector.
+     */
+    struct {
+      /**
+       * @brief   PIO block callback function.
+       */
+      rp_pioisr_t   func;
+      /**
+       * @brief   PIO block callback parameter.
+       */
+      void          *param;
+    } block;
+    /**
      * @brief   PIO state machine IRQ redirectors.
      */
     struct {
@@ -137,6 +150,12 @@ static void serve_pio_irq(uint32_t blockidx, __I uint32_t *ints_reg) {
 
   if (ints != 0U) {
     unsigned i;
+
+    /* The block callback is invoked once, before the per state machine
+       ones, so it can acknowledge the source on their behalf.*/
+    if (pio.blocks[blockidx].block.func != NULL) {
+      pio.blocks[blockidx].block.func(pio.blocks[blockidx].block.param, ints);
+    }
 
     for (i = 0U; i < RP_PIO_NUM_STATE_MACHINES; i++) {
       if (pio.blocks[blockidx].sm[i].func != NULL) {
@@ -240,6 +259,7 @@ void pioInit(void) {
     pio.blocks[b].c0_allocated_mask = 0U;
     pio.blocks[b].c1_allocated_mask = 0U;
     pio.blocks[b].imem_allocated    = 0U;
+    pio.blocks[b].block.func        = NULL;
     for (s = 0U; s < RP_PIO_NUM_STATE_MACHINES; s++) {
       pio.blocks[b].sm[s].func = NULL;
     }
@@ -702,6 +722,51 @@ void pioSmInit(const rp_pio_sm_t *smp, uint32_t initial_pc,
   pioSmRestartX(smp);
   pioSmClkdivRestartX(smp);
   pioSmSetPCX(smp, initial_pc);
+}
+
+/**
+ * @brief   Associates a callback to a PIO block.
+ * @details The callback is invoked once per interrupt of the block, before
+ *          the callbacks of the allocated state machines, with the content
+ *          of the IRQn_INTS register.
+ * @note    The driver acknowledges nothing itself: IRQn_INTS is derived
+ *          from INTR and the FIFO levels, so unless the source is cleared
+ *          the interrupt fires again immediately. A block callback is the
+ *          natural place to do it, being the only handler guaranteed to
+ *          run exactly once per interrupt.
+ * @note    Passing @p NULL removes the callback.
+ *
+ * @param[in] block     pointer to the PIO block descriptor
+ * @param[in] func      callback function, can be @p NULL
+ * @param[in] param     parameter passed to the callback
+ *
+ * @iclass
+ */
+void pioSetBlockCallbackI(const rp_pio_block_t *block,
+                          rp_pioisr_t func, void *param) {
+
+  osalDbgCheckClassI();
+  osalDbgCheck(block != NULL);
+
+  pio.blocks[block->pioidx].block.param = param;
+  pio.blocks[block->pioidx].block.func  = func;
+}
+
+/**
+ * @brief   Associates a callback to a PIO block.
+ *
+ * @param[in] block     pointer to the PIO block descriptor
+ * @param[in] func      callback function, can be @p NULL
+ * @param[in] param     parameter passed to the callback
+ *
+ * @api
+ */
+void pioSetBlockCallback(const rp_pio_block_t *block,
+                         rp_pioisr_t func, void *param) {
+
+  osalSysLock();
+  pioSetBlockCallbackI(block, func, param);
+  osalSysUnlock();
 }
 
 /**
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.c
@@ -511,7 +511,8 @@ void pioSmFreeI(const rp_pio_sm_t *smp) {
   if (((pio.blocks[b].c0_allocated_mask |
         pio.blocks[b].c1_allocated_mask) == 0U) &&
       (pio.blocks[b].imem_allocated == 0U)) {
-    pio.blocks[b].block.func = NULL;
+    pio.blocks[b].block.func  = NULL;
+    pio.blocks[b].block.param = NULL;
     rp_peripheral_reset(smp->block->resets_mask);
   }
 }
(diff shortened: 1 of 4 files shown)


chibios-sheriff (commented):

Scope: RP PIOv1 — add block-level interrupt enable/disable, a block-level IRQ callback, and block-based GPIO routing primitives. Base `master`.

> not the full review yet.

Nice work — well-reasoned, self-documented, and it comes with validation coverage (cases 15–17). Everything checks out:

- **Registers (verified vs RP2040 datasheet).** `pioEnableInterruptX()` targets `IRQ0_INTE`/`IRQ1_INTE`, which are exactly the two per-request-line interrupt-enable registers (§3.5.7 "the lower 4 can be masked into one of PIO's interrupt request lines, via the IRQ0_INTE and IRQ1_INTE control registers"), and the doxygen's INTE bit layout (flags 0–3 at bits 8–11 on RP2040, 0–7 at 8–15 on RP2350) matches. The `CPUID → IRQ0/IRQ1` mapping is consistent with the existing `pioSmEnableInterruptX()`.
- **Block callback.** Invoked once per block interrupt, before the per-SM callbacks, with the full `IRQn_INTS` word — the right place for the once-per-interrupt acknowledge, and the "driver acks nothing itself" contract is clearly documented.
- **Backward-compatible refactor.** The GPIO routing is factored into block-level primitives (`pioSetPinFunctionX`, `pioGpioRoutePadX`, `pioGpioRouteX`) while the SM-based entry points (`pioGpioInitX`, `pioGpioInitPadX`, `pioSmSetPinFunctionX`) are kept as thin `smp->block` delegators — no API break (no orphaned callers in the tree).
- **Concurrency.** `pioSetBlockCallbackI` is `@iclass` with `osalDbgCheckClassI()`, the public wrapper locks; the `X` interrupt helpers use the atomic `SET`/`CLR` aliases, so lock-free is fine. `pioDisableInterruptX` clearing both cores' INTE is documented and safe.

Mechanical gates: `stylecheck.py` clean on both files; `testhal/RP/{RP2040,RP2350}/PIO-VALIDATION` both build clean, ARM GCC 14.3.1.

No changelog entry needed on its own (the RP PIO API build-out is unreleased and covered by the broader XHAL/RP work); flag me if you'd prefer one.

emolitor:
Will review this evening.

emolitor:
Reviewed. The block-level API is the right shape and the Sm-handle refactor is a clean pure-forward, but I think the central premise needs an answer before this goes to the bench.

**Mechanical state** (checked locally against master at `6b3d5c8b6e`): merges clean, and the RP2350 Cortex-M33 PIO-VALIDATION leg builds clean here — only the pre-existing `LOAD segment with RWX permissions` linker warning. Against its three siblings the only conflicts are the shared test insertion point in the two `main.c` files, no code conflicts in `rp_pio.h`.

**The "no state machine allocated" premise does not hold.** Three separate reasons, all from the current driver:

1. An idle PIO block is *held in reset*. `pioSmAllocI()` and `pioProgramLoad()` unreset it; `pioSmFreeI()` and `pioProgramUnloadI()` reset it again once the block is fully idle. `pioEnableInterruptX()` writes INTE with no unreset in front of it. The driver already documents exactly this hazard for the sibling case — `rp_pio.c:806`, in `pioSetGpioBase()`:

   > An idle block is held in reset, it must be released for the GPIOBASE write to take effect. Resetting it back would clear the register so the block is left out of reset.

   The same applies to IRQ0_INTE/IRQ1_INTE. On a genuinely cold block the enable is lost.

2. The PIO NVIC vector is enabled *only* in `pioSmAllocI()` (`rp_pio.c:309-337`) and disabled in `pioSmFreeI()` when a core's last state machine is released (`rp_pio.c:444-472`). So even with INTE set, no interrupt reaches the CPU unless some state machine is allocated on this core. A block interrupt enabled with nothing allocated cannot fire.

3. Nothing clears the block callback on teardown. `pioSmFreeI()` clears the per-state-machine callback and resets the block when it goes fully idle, but `block.func`/`block.param` survive — a registered block callback silently stops working (INTE wiped by the reset, vector disabled) while the driver still holds the pointer and the parameter.

**The suite would not catch (1).** Tracing the RP2350 `main.c` as it stands after this branch: the last `pioProgramUnload()` before this region is at line 872, and test 14 ends with `pioSmFree(smp)`, so both allocation masks and `imem_allocated` are zero and the block is back in reset when test 15 starts. `report("INTE idle on entry", ...)` then reads a block in reset, and `report("enable reaches this core's INTE", block->pio->IRQ0_INTE == PIO_IRQ_SM(0))` should fail because the write never landed. Test 16 sidesteps (2) by allocating SM0 before registering the block callback, so it exercises the working path rather than the documented one. This is a static trace, not a bench run — but it is why I would rather settle the design than spend a bench session discovering it.

**Two directions, your call:**

- *Cheap:* scope the documentation down to a block that is already active (at least one state machine allocated or one program loaded), drop the "no state machine has to be allocated to reach it" claim, and rework test 15 to allocate first. `pioEnableInterruptX()`/`pioDisableInterruptX()` stay X-class inlines and stay useful — the value is being able to reach the block INTE without holding a specific SM handle, which survives intact.
- *Bigger:* give the block an acquire/release pair that unresets, enables the vector, and clears the callback on release. That makes the stated premise true, but it is a new lifecycle concept in the driver and the tests get rewritten.

Everything else here reads well: the GPIO routing consolidation is a genuine improvement (one home for the RP2350 isolation sequence), the three Sm forwarders are byte-for-byte equivalent, and the NULL fast path in `serve_pio_irq()` is the right cost. On the naming question you raised — `pioGpioRoute*X` for the block versions with the Sm names untouched is the right trade; renaming the Sm variants would break existing callers for cosmetics.

emolitor (changes requested):
See comment about mismatch between no state machines required. This might be as small as a document change but should be consistent.

begeistert:
Review taken — thanks, the reset trace was exactly right.

I went with the cheap direction, plus your point 3 for consistency:

- `pioEnableInterruptX()` no longer claims no allocation is needed: `@pre` the block is active (an idle block is held in reset and loses the INTE write), and a `@note` ties the vector lifecycle to the state machine allocation.
- `pioSmFreeI()` now drops the block callback when the block becomes fully idle and is reset — a registration that can never fire again is a stale pointer, and this mirrors the per state machine callbacks being cleared on free.
- Test 15 allocates SM0 first (with an acknowledging callback so the enabled vector cannot storm) and now validates the full flag -> INTS -> ISR path through the block enable, plus the pending-but-masked state after the disable.

The branch is rebased onto current master (tests 15..17 sit ahead of the encoders test from #225) with the review changes as a separate commit on top.

Bench results on a Pico 2 (RP2350), report captured from UART0: **139 pass, 0 fail**. The Test 15 sub-case also settles the IRQ_FORCE question from the PR description: a forced flag reads back through `pioIrqGetX()` and clears through `pioIrqClearX()` exactly as their doc states, so no separate accessor is needed. RP2040 leg run on a Pico W: **123 pass, 0 fail** — both legs green.

begeistert:
Working on the merge conflicts! 😄

emolitor:
Reviewed the follow-up and benched it. The design fix is right on all three points — the `@pre`/`@note` scoping is honest about what the API can and cannot do, dropping the callback in `pioSmFreeI()` is the correct lifetime rule, and the reworked test 15 is a better test than the original: asserting `forced flag reaches the ISR` exercises the whole flag → INTS → ISR path, where the old version only proved a register had been written.

**Bench record**, run here on a cross-wired RP2040 + RP2350 pair, all three legs:

| Leg | Result |
| --- | --- |
| RP2040 (Cortex-M0+) | 123 pass, 0 fail |
| RP2350 (Cortex-M33) | 139 pass, 0 fail |
| RP2350 RISC-V (Hazard3) | 139 pass, 0 fail |

The first two reproduce your numbers exactly. The RISC-V leg was not in your report — green too.

## One gap left: the other teardown door

`pioSmFreeI()` now drops the block callback, but `pioProgramUnloadI()` resets the block on the same condition without dropping it:

```c
  /* Reset PIO block if it became fully idle, no state machines allocated
     by either core and no programs loaded.*/
  if (((pio.blocks[b].c0_allocated_mask |
        pio.blocks[b].c1_allocated_mask) == 0U) &&
      (pio.blocks[b].imem_allocated == 0U)) {
    rp_peripheral_reset(block->resets_mask);   /* block.func survives */
  }
```

Reachable ordering: allocate a state machine → register a block callback → load a program → free the state machine (the block stays alive because the program still holds it, so the callback is correctly retained) → unload the program. Now the block goes fully idle through the *unload* path, gets reset, INTE is wiped, and `block.func` survives. That is the same stale registration the new commit exists to prevent, just reached through the other door.

Suggested change, mirroring what you already did in `pioSmFreeI()`:

```diff
-  /* Reset PIO block if it became fully idle, no state machines allocated
-     by either core and no programs loaded.*/
+  /* Reset PIO block if it became fully idle, no state machines allocated
+     by either core and no programs loaded. The reset wipes the INTE
+     routing, so the block callback is dropped with it, as in
+     pioSmFreeI().*/
   if (((pio.blocks[b].c0_allocated_mask |
         pio.blocks[b].c1_allocated_mask) == 0U) &&
       (pio.blocks[b].imem_allocated == 0U)) {
+    pio.blocks[b].block.func = NULL;
     rp_peripheral_reset(block->resets_mask);
   }
```

Minor, same area: both sites clear `block.func` but leave `block.param` set, where the per state machine path two lines above clears `func` and `param` together. Harmless since `func == NULL` gates the call, but the asymmetry invites a second look later — worth clearing both in each place.

## A test that would catch it

The fix is internal state, so it needs a behavioural probe. Reallocating after the unload makes it observable:

1. allocate SM0 with an acknowledging callback, register a block callback, load a program;
2. free SM0, then unload the program — the block goes idle through the unload path;
3. allocate SM0 again, `pioEnableInterruptX()`, `pioIrqForceX()`;
4. assert the block callback did **not** run.

Without the fix step 3 re-enables the vector and INTE, `serve_pio_irq()` still sees a non-NULL `block.func`, and the stale callback fires — so the assert separates the two states cleanly. It also pins the intended rule, which is that a block callback does not survive the block going idle by any route.

Nothing else outstanding from my side; the API shape, the GPIO routing consolidation and the callback ordering all look right to me.

emolitor (changes requested):
One more round of small changes to consider then we can land this.

begeistert:
Good catch — fixed, and your probe is now part of test 16.

- `pioProgramUnloadI()` drops the block callback on its reset path, mirroring `pioSmFreeI()`, and both teardown sites now clear `func` and `param` together, matching the per state machine path (the asymmetry nit included).
- Test 16 grew exactly the probe you outlined: idle the block through the unload path, reallocate, force a flag, assert the old callback stays silent. It pins the rule that a block callback does not survive the block going idle by any route.

Bench with the fix as the tip commit, reports captured from UART0:

| Leg | Result |
| --- | --- |
| RP2040 (Pico W) | 147 pass, 0 fail |
| RP2350 (Pico 2) | 165 pass, 0 fail |

The totals grew over my previous run because the rebased base now carries the merged tests 18..21 too; the new `unload-path teardown dropped the callback` assert is green on both legs. Thanks for running the RISC-V leg — I owe that one a spot on my bench.

The branch is rebased onto current master (post-#223) and mergeable; nothing else pending from my side.

emolitor (approved):

PR #223, opened 2026-08-11
RP PIOv1: add runtime state machine setters and readback

begeistert:
## Summary

`pioSmSetExecctrlX()` and `pioSmSetPinctrlX()` write their whole register, so changing one field of a configured state machine at runtime clears every other one. Repointing the wrap range at a second program — the way a single state machine is reused for several programs — through the full write destroys JMP_PIN, SIDE_EN, SIDE_PINDIR, OUT_STICKY, INLINE_OUT_EN and OUT_EN_SEL, so consumers open-code the read-modify-write (the nanoFramework CYW43 SPI driver carries exactly that, with a comment noting the full-write hazard). In the other direction, CLKDIV, GPIOBASE and INPUT_SYNC_BYPASS had setters or nothing but no readback, and the nanoFramework PIO binding (nanoframework/nf-interpreter#3449) reads the raw registers today.

This PR:

- adds `pioSmSetWrapX()` and `pioSmSetJmpPinX()`, modifying only their EXECCTRL fields (`pio_sm_set_wrap()` in the pico-sdk);
- adds `pioSmSetOutPinsX()`, `pioSmSetSetPinsX()`, `pioSmSetInPinBaseX()` and `pioSmSetSidesetPinsX()`, the runtime mirrors of the `pioSmConfigSet*` builders on PINCTRL, with the same field validation. SIDESET_COUNT deliberately has no runtime setter: changing the side-set width under a loaded program is incoherent, moving the base is not;
- adds `pioSmGetClkdivX()` (raw CLKDIV, decomposable with the existing `PIO_SM_CLKDIV_*` macros) and, on the RP2350, `pioGetGpioBaseX()`, the dual of `pioSetGpioBase()` under the same `RP_PIO_HAS_GPIOBASE` gate;
- adds `pioSetInputSyncBypassX()`/`pioGetInputSyncBypassX()` over the atomic SET/CLR aliases (`pio_set_input_sync_bypass()` in the pico-sdk), saving the two clock cycles of input synchronizer latency for signals synchronous to the system clock; the doc notes the RP2350 window semantics (bit n is GPIO base+n).

PIO-VALIDATION grows tests 18..19: test 18 plants distinctive values in every neighbouring EXECCTRL/PINCTRL field, touches each setter and checks the whole register bit-exact (the field without a setter must survive); test 19 does setter/getter roundtrips for CLKDIV, the sync bypass and (RP2350) GPIOBASE — the latter after the state machine free, since the pin window only moves on an idle block. Test numbers 15..22 are reserved across this PIO API series; this branch carries only its own tests.

## Related

- Issue(s): —
- Notes for reviewers: pin numbers in the new setters are window-relative on the RP2350, documented via `pioGpioToRel()` like the rest of the driver. Independent of the other PRs in the series (#222 block level APIs, plus patching/synchronized enable and instruction encoders to follow).

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

## Testing

- RP2350 leg run on a Pico 2: **124 pass, 0 fail**; RP2040 leg run on a Pico W: **107 pass, 0 fail** (reports captured from UART0).

## Compatibility and Risk

Additive only: no existing function changes name, signature or behavior. The new setters are read-modify-write and therefore not atomic against a concurrent full-register write from the other core; that is the same caveat every RMW field access on these registers carries and is not introduced by this PR.

### DIFF
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
@@ -1099,22 +1099,22 @@ __STATIC_INLINE void pioSmSetPinctrlX(const rp_pio_sm_t *smp,
  *          @p RP_PIO_HAS_GPIOBASE capability, see @p pioGpioToRel().
  *
  * @param[in] smp       pointer to a rp_pio_sm_t structure
- * @param[in] base      first pin of the OUT group (0..31)
+ * @param[in] pin_base  first pin of the OUT group (0..31)
  * @param[in] count     number of pins in the OUT group (0..32)
  *
  * @special
  */
 __STATIC_INLINE void pioSmSetOutPinsX(const rp_pio_sm_t *smp,
-                                      uint32_t base, uint32_t count) {
+                                      uint32_t pin_base, uint32_t count) {
   uint32_t pinctrl;
 
-  osalDbgCheck((base < 32U) && (count <= 32U));
+  osalDbgCheck((pin_base < 32U) && (count <= 32U));
 
   pinctrl = smp->block->pio->SM[smp->smidx].PINCTRL;
   smp->block->pio->SM[smp->smidx].PINCTRL =
     (pinctrl & ~(PIO_SM_PINCTRL_OUT_BASE_Msk |
                  PIO_SM_PINCTRL_OUT_COUNT_Msk)) |
-    (base << PIO_SM_PINCTRL_OUT_BASE_Pos) |
+    (pin_base << PIO_SM_PINCTRL_OUT_BASE_Pos) |
     (count << PIO_SM_PINCTRL_OUT_COUNT_Pos);
 }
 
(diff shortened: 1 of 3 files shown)


chibios-sheriff (commented):

Scope: RP PIOv1 — runtime state-machine setters/readback (CLKDIV, wrap, jmp-pin, OUT/SET/IN/sideset pin fields), `INPUT_SYNC_BYPASS` control, and `GPIOBASE` readback. Base `master`.

> not the full review yet.

Clean and consistent with the existing driver style. The field setters are proper masked read-modify-writes on `EXECCTRL`/`PINCTRL` using the CMSIS `PIO_SM_*` bitfield macros, and the debug-assert bounds match the hardware field limits — `SET count ≤ 5`, `OUT count ≤ 32`, wrap targets `< RP_PIO_NUM_INSTR_MEM`, pin bases `< 32`. `INPUT_SYNC_BYPASS` uses the atomic `SET`/`CLR` aliases, and the `GPIOBASE` readback is correctly gated behind `RP_PIO_HAS_GPIOBASE` (RP2350). Test 18 exercises the new setters.

emolitor:
Will review this evening.

emolitor:
Reviewed. This is the most mechanical of the four and I found nothing substantive to fix.

**The setters check out against their config counterparts.** I compared each new runtime setter with the builder it mirrors and they match field for field — same masks, same bounds, same read-modify-write shape:

- `pioSmSetWrapX` / `pioSmConfigSetWrapX` (`rp_pio.h:512`)
- `pioSmSetJmpPinX` / `pioSmConfigSetJmpPinX` (`rp_pio.h:633`)
- `pioSmSetOutPinsX` / `pioSmConfigSetOutPinsX` — `base < 32`, `count <= 32` both sides
- `pioSmSetSetPinsX` / `pioSmConfigSetSetPinsX` — `count <= 5` both sides, correct for the 3-bit SET_COUNT
- `pioSmSetInPinBaseX` / `pioSmConfigSetInPinsX`
- `pioSmSetSidesetPinsX` / `pioSmConfigSetSidesetPinsX`

Leaving SIDESET_COUNT without a runtime setter is the right call for the reason you give.

`pioSetInputSyncBypassX()` over the SET/CLR aliases is exactly right — atomic, no read, consistent with the rest of the driver.

**Naming nits, take or leave:**

- The existing builders call the parameter `pin_base`; the new setters call it `base`. Worth matching the builders since they sit in the same header and document the same window-relative rule.
- `pioSmSetInPinBaseX` vs. its stated counterpart `pioSmConfigSetInPinsX` — the asymmetry is defensible (it really does only set the base) but it is the one name in the group that does not read as the runtime twin of the builder above it.

**One caveat on `pioGetGpioBaseX()`.** It reads GPIOBASE on a block that may be held in reset — an idle block is, and `pioSetGpioBase()` deliberately leaves the block *out* of reset for precisely this reason (`rp_pio.c:806`). In practice this is harmless: a block that has never had its base moved reads back 0, which is also the correct answer. Not worth code, but a `@note` that the value is only meaningful on a block that has been configured or is active would save someone the confusion. Test 19 already does the right thing by reading it after the state machine free rather than on a cold block.

emolitor (changes requested):
This is good to go once you resolve the merge conflicts.

begeistert:
Conflicts resolved: rebased onto current master, with tests 18..19 ordered ahead of the encoders test from #225 and the pin setters ahead of the encoders group in `rp_pio.h`. The take-or-leave nits were worth taking, as a separate commit: the setters take `pin_base` like the builders they mirror, `pioSmSetInPinBaseX()` is now `pioSmSetInPinsX()` (cheap today, a breaking rename after the merge), and `pioGetGpioBaseX()` carries your note about an idle block reading back 0.

Bench results on a Pico 2 (RP2350), report captured from UART0: **124 pass, 0 fail** — the bit-exact RMW checks and the CLKDIV/GPIOBASE/sync-bypass roundtrips all green. RP2040 leg run on a Pico W: **107 pass, 0 fail** — both legs green.

begeistert:
Working on the merge conflicts! 😄

emolitor:
Pushed the one-word doc fix to this branch directly (maintainer edit) rather than round-tripping it — `638e4216eb`. The `pin_base` rename substitution had caught the `@brief` line of `pioSmSetSidesetPinsX()`, leaving "Sets the side-set pin **pin_base** of a running state machine". No code change; style check clean.

Bench record for this branch, run here on a cross-wired RP2040 + RP2350 pair, all three legs:

| Leg | Result |
| --- | --- |
| RP2040 (Cortex-M0+) | 107 pass, 0 fail |
| RP2350 (Cortex-M33) | 124 pass, 0 fail |
| RP2350 RISC-V (Hazard3) | 124 pass, 0 fail |

The RP2040 and RP2350 ARM legs reproduce your numbers exactly. The RISC-V leg was not in your report — it is green too, and I confirmed the chip really was executing Hazard3 for that run (`rp2350.rv0 running`, `rp2350.cm0 unavailable`) rather than falling back to ARM. Tests 18 and 19 pass on all three: wrap and JMP_PIN repointed with neighbouring fields preserved, PINCTRL rebuilt field by field, and the CLKDIV / sync-bypass / GPIOBASE roundtrips.

Those runs were on `c3a32b5e` and this branch has since been rebased past #224, so I re-checked: the functions this PR adds are byte-identical across the rebase, so the results carry over unchanged.

Nothing else outstanding from my side.

emolitor (approved):

PR #224, opened 2026-08-11
RP PIOv1: add instruction patching and synchronized enable

begeistert:
## Summary

Two program-lifecycle operations had no API and forced consumers onto the raw registers:

- **`pioProgramPatchX(block, addr, instr)`** rewrites one instruction memory slot of a loaded program. The loader already relocates JMP targets *at load time*; this covers relocation *after* the load — repointing a JMP at runtime, which the nanoFramework CYW43 PIO-SPI driver does today by writing `pio->INSTR_MEM[offset + i]` directly. Addresses follow the same rebasing rule as the loader (both `addr` and any JMP target encoded in `instr` are absolute, offset included), and the doc says so.
- **`pioEnableSmMaskInSyncX(block, mask)`** raises `SM_ENABLE` and `CLKDIV_RESTART` for a mask of state machines in a single CTRL read-modify-write, so their divided clocks restart together and the machines run cycle-aligned. There is no way to do this with the existing per-machine calls: `pioSmEnableX()` + `pioSmClkdivRestartX()` are separate CTRL writes and with dividers > 1 the machines come up with arbitrary relative phase. The RMW (rather than a full register write) preserves the enables of machines outside the mask; the CLKDIV_RESTART bits are self-clearing and read back as zero, so they cannot pollute the OR. This is `pio_enable_sm_mask_in_sync()` in the pico-sdk.

PIO-VALIDATION grows tests 20..21. The instruction memory is write-only on both chips, so test 20 is behavioral rather than a readback: the parking program pins address 0 so the square wave loads at a non-zero offset (exercising the rebasing), then a live patch of `set pins, 1` to `set pins, 0` freezes the pin without stopping the machine, and the patch-back restores the wave. Test 21 checks both enables appear with one CTRL write and then samples `SIO->GPIO_IN` for 100 ms requiring the two pins driven by the same program on SM0/SM1 to be equal in every single read — a phase misalignment would make ~2/3 of the samples differ, so the pass/fail separation is wide and the test demands zero mismatches. Test numbers 15..22 are reserved across this PIO API series; this branch carries only its own tests.

## Related

- Issue(s): —
- Notes for reviewers — two deliberate design choices here are the weak points and I would value your call on both:
  1. **`pioProgramPatchX()` does not assert slot ownership.** Validating `addr` against the instruction memory allocation bitmap would require either calling `pioGetImemAllocatedMask()` (an `@api` function that takes the system lock — not callable from an X-class inline) or moving the function into `rp_pio.c` behind the lock, which taxes every caller for a setup-path operation. Ownership is a documented precondition instead ("the caller must own the slot through a previous `pioProgramLoad()`"). If you prefer the locked `.c` variant, or an I-class flavor with a debug assert, happy to rework.
  2. **No `pio_set_sm_mask_enabled()` counterpart.** Only the in-sync enable is added; a mask enable/disable *without* the divider restart was left out to keep the PR minimal. It is a two-line addition if you want the pair completed here.

  Independent of the other PRs in the series (#222 block level APIs, #223 runtime setters/readback, instruction encoders to follow).

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

## Testing

- RP2350 leg run on a Pico 2: **123 pass, 0 fail**; RP2040 leg run on a Pico W: **108 pass, 0 fail** (reports captured from UART0).

## Compatibility and Risk

Additive only. `pioProgramPatchX()` trusts the caller on slot ownership (see reviewer note 1); a patch to a slot owned by another program is a documented misuse the debug build cannot currently catch. `pioEnableSmMaskInSyncX()` is a read-modify-write on CTRL and therefore not atomic against a concurrent CTRL write from the other core — the same caveat every RMW on a shared block register carries.

### DIFF
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
@@ -922,10 +922,11 @@ __STATIC_INLINE void pioSmDisableX(const rp_pio_sm_t *smp) {
 
 /**
  * @brief   Enables several state machines of a block simultaneously.
- * @details A single CTRL read-modify-write raises the SM_ENABLE and
- *          CLKDIV_RESTART bits of every state machine in the mask, so
- *          their clock dividers restart together and the machines run
- *          cycle-aligned. Machines outside the mask are not affected.
+ * @details A single write to the atomic SET alias raises the SM_ENABLE
+ *          and CLKDIV_RESTART bits of every state machine in the mask,
+ *          so their clock dividers restart together and the machines run
+ *          cycle-aligned. Machines outside the mask are not affected by
+ *          construction, and the write is atomic against the other core.
  * @pre     Each state machine in the mask is configured and its program
  *          counter is positioned, as before an individual
  *          @p pioSmEnableX().
@@ -937,12 +938,10 @@ __STATIC_INLINE void pioSmDisableX(const rp_pio_sm_t *smp) {
  */
 __STATIC_INLINE void pioEnableSmMaskInSyncX(const rp_pio_block_t *block,
                                             uint32_t mask) {
-  uint32_t ctrl;
 
   osalDbgCheck((block != NULL) && ((mask & ~0xFU) == 0U));
 
-  ctrl = block->pio->CTRL;
-  block->pio->CTRL = ctrl | mask | (mask << 8U);
+  block->pio->SET.CTRL = mask | (mask << 8U);
 }
 
 /**
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
@@ -921,6 +921,31 @@ __STATIC_INLINE void pioSmDisableX(const rp_pio_sm_t *smp) {
 }
 
 /**
+ * @brief   Enables several state machines of a block simultaneously.
+ * @details A single CTRL read-modify-write raises the SM_ENABLE and
+ *          CLKDIV_RESTART bits of every state machine in the mask, so
+ *          their clock dividers restart together and the machines run
+ *          cycle-aligned. Machines outside the mask are not affected.
+ * @pre     Each state machine in the mask is configured and its program
+ *          counter is positioned, as before an individual
+ *          @p pioSmEnableX().
+ *
+ * @param[in] block     pointer to the PIO block descriptor
+ * @param[in] mask      mask of state machines, bits 0..3
+ *
+ * @special
+ */
+__STATIC_INLINE void pioEnableSmMaskInSyncX(const rp_pio_block_t *block,
+                                            uint32_t mask) {
+  uint32_t ctrl;
+
+  osalDbgCheck((block != NULL) && ((mask & ~0xFU) == 0U));
+
+  ctrl = block->pio->CTRL;
+  block->pio->CTRL = ctrl | mask | (mask << 8U);
+}
+
+/**
  * @brief   Restarts a state machine.
  *
  * @param[in] smp       pointer to a rp_pio_sm_t structure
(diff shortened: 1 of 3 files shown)


chibios-sheriff (commented):

Scope: RP PIOv1 — in-place instruction patching (`pioProgramPatchX`) and synchronized multi-SM enable (`pioEnableSmMaskInSyncX`). Base `master`.

> not the full review yet.

Both correct against the hardware:

- **`pioEnableSmMaskInSyncX`** writes `CTRL |= mask | (mask << 8)` — `SM_ENABLE[3:0]` plus `CLKDIV_RESTART[11:8]` (bit positions confirmed in the RP2040 datasheet) in a single write, so the selected machines start with their dividers restarted together and run cycle-aligned. This matches the pico-SDK `pio_enable_sm_mask_in_sync` technique. The `mask & ~0xF == 0` check keeps it to the four SM bits.
- **`pioProgramPatchX`** writes a 16-bit instruction into `INSTR_MEM[addr]` with `addr < RP_PIO_NUM_INSTR_MEM`.

The validation is genuinely thorough: Test 20 patches a running `set pins,1` → `set pins,0` and back and checks the pin freezes/resumes without stopping the machine; Test 21 launches two SMs in sync on adjacent pins and confirms they stay locked in phase (a single `GPIO_IN` read always sees the two pins equal).

emolitor:
Will review this evening.

emolitor:
Reviewed. Both operations are real gaps and the tests are well designed — test 21's "zero mismatches over 100 ms" gives a genuinely wide pass/fail separation. One concrete change I would like before this goes to the bench.

**`pioEnableSmMaskInSyncX()` should use the atomic SET alias, not a read-modify-write.** The bits you want are all write-1-to-set: SM_ENABLE is set-only here (you never clear), and CLKDIV_RESTART is self-clearing. So:

```c
block->pio->SET.CTRL = mask | (mask << 8U);
```

One write instead of a read plus a write, atomic against the other core, and it preserves machines outside the mask by construction rather than by having read them first. It also matches what the driver already does everywhere else on CTRL — `pioSmEnableX()` at `rp_pio.h:854`, `pioSmRestartX()` at `rp_pio.h:878`, `pioSmClkdivRestartX()` at `rp_pio.h:890` all go through `SET.CTRL`.

This removes the caveat in your Compatibility and Risk section rather than documenting it. The simultaneity argument is unaffected: it is still a single write to CTRL, so the enables and the divider restarts land together exactly as they do now. Worth changing before the bench run so you validate the final code — and `report("both enables set by one write", (block->pio->CTRL & 0xFU) == 0x3U)` still holds as written.

**On your reviewer note 1 (`pioProgramPatchX()` slot ownership):** I would keep it as you have it. Documented precondition, X-class inline, no lock. Taking the system lock on a setup-path operation to validate against `imem_allocated` would be a real cost for a misuse the caller is already in a position to avoid, and moving it into `rp_pio.c` would make it the only instruction-memory operation that cannot be used from a fast path. The rebasing rule being identical to the loader's is the part that matters, and the doc states it.

**On your reviewer note 2 (no `pio_set_sm_mask_enabled()` counterpart):** agreed, leave it out. The in-sync enable exists because there is no way to express it with the per-machine calls; a plain mask enable is just `pioSmEnableX()` in a loop and adds an API for nothing.

Test 20 is a nice piece of work — pinning address 0 with the parking program so the square wave has to load at a non-zero offset means the rebasing is actually exercised rather than assumed, which matters given the instruction memory is write-only and cannot be checked directly.

emolitor (changes requested):
Minimal change requested please use the atomic SET alias, not a read-modify-write

begeistert:
Done — `pioEnableSmMaskInSyncX()` now writes `SET.CTRL = mask | (mask << 8U)`, as a separate commit on top of the rebase, and the concurrency caveat in the Compatibility section is gone with the read-modify-write. Notes 1 and 2 stay as discussed.

Bench results on a Pico 2 (RP2350), report captured from UART0: **123 pass, 0 fail** — test 21's phase lock (zero mismatches over the 100 ms window) passed with the SET-alias version, so what got validated is the final code. RP2040 leg run on a Pico W: **108 pass, 0 fail** — both legs green.

emolitor (approved):

PR #225, opened 2026-08-11
RP PIOv1: add instruction encoders

begeistert:
## Summary

`pioSmExecX()` and `rp_pio_program_t` instruction arrays take raw instruction words and the driver gives no way to build one, so every consumer hand-assembles literals: `0xE081`, `0x6020`, `0xC003` — the PIO-VALIDATION suite is full of them, downstream drivers carry their own (the nanoFramework CYW43 PIO-SPI driver, the PIO binding in nanoframework/nf-interpreter#3449), and the driver itself was doing it in two places. Hand-assembly is unreadable and easy to get wrong; the `PULL`-encoded-as-`PUSH` bug fixed during the TX FIFO drain review (#168) is exactly the class of mistake an encoder makes hard to write.

This PR:

- adds operand macros: `PIO_DEST_*`, `PIO_SRC_*`, `PIO_MOV_*`, all eight `PIO_JMP_*` conditions, `PIO_WAIT_*` sources (`PIO_WAIT_JMPPIN` gated to the RP2350, where it exists), and `PIO_IRQ_INDEX_REL` for machine-relative flag addressing. EXEC is deliberately not a single constant: OUT encodes it as 7 and MOV as 4, so `PIO_OUT_DEST_EXEC` and `PIO_MOV_DEST_EXEC` are separate and cannot be crossed;
- adds `pioEncodeJmp/Wait/In/Out/Push/Pull/Mov/Irq/Set/Nop()`, `__STATIC_INLINE` with `osalDbgCheck`ed operand ranges. The checks catch the two classic encoding traps: IN/OUT bit count 32 is encoded as 0 (validated as 1..32, emitted masked), and MOV op 3 is reserved (rejected);
- adds the delay/side-set field combinators `pioEncodeDelay()` and `pioEncodeSideSet()`, ORed into any instruction. The delay field shrinks as side-set bits are configured, so both take the configured side-set width and validate the value fits instead of silently corrupting the neighbouring field; `pioEncodeSideSet()` handles the optional (SIDE_EN) enable bit placement;
- converts the driver's two internal hand-assembled sites: `pioSmSetPCX()` now emits `pioEncodeJmp(PIO_JMP_ALWAYS, ...)` and `pioSmSetConsecutivePindirsX()` emits `pioEncodeSet(PIO_DEST_PINDIRS, ...)`. The emitted instructions are bit-identical, and the existing `PIO_INSTR_*` constants are untouched.

Functionally this is `pio_encode_*` in the pico-sdk; the encodings follow the datasheet instruction set tables (RP2040/RP2350 §3.4).

PIO-VALIDATION grows test 22: seventeen asserts checking every encoder against the hand-assembled words the suite and the driver already use (`pioEncodeSet(PIO_DEST_PINDIRS, 1) == 0xE081`, `pioEncodeNop() == PIO_INSTR_NOP`, `pioEncodePull(false, false) == PIO_INSTR_PULL_NOBLOCK`, ...), pure computation that gives full signal without a board, plus a behavioral leg: the reference square wave rebuilt from encoders alone, verified word-identical to the literal program and then run on the pin. The RP2350 leg adds a `PIO_WAIT_JMPPIN` encoding check. Test numbers 15..22 are reserved across this PIO API series; this branch carries only its own test.

## Related

- Issue(s): —
- Notes for reviewers: two conventions worth your call. (1) The encoders carry no X suffix — they are pure functions with no register effect, following the precedent of `PIO_INSTR_NOP` and `PIO_SM_CLKDIV()`; a rename is mechanical if you prefer X. (2) `pioEncodeIrq()` takes the flag index with `PIO_IRQ_INDEX_REL` ORed in (checked < 32) rather than a separate `bool rel` parameter, so the signature survives the RP2350 index-mode extensions unchanged. Last PR of the series (#222 block level APIs, #223 runtime setters/readback, #224 patching/synchronized enable), independent of the other three.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

## Testing

- The three PIO-VALIDATION legs compile clean; the device run of the suite is pending on Pico/Pico2 hardware and results will be reported in this thread. Most of test 22 is pure computation (encoder words vs. known literals) and gives full signal on any run; the encoder-built square wave leg needs the board.

## Compatibility and Risk

Additive, plus two internal call sites converted with bit-identical emitted instructions. The encoders validate operand ranges, not per-instruction operand legality (e.g. `SET` accepts any `PIO_DEST_*` value in range even though the hardware only implements four destinations); full legality tables are documented as the datasheet's domain, matching the existing `PIO_DEST_*` doc note.

### DIFF
--- a/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
+++ b/os/hal/ports/RP/LLD/PIOv1/rp_pio.h
@@ -257,6 +257,60 @@
 #define PIO_INSTR_NOP                   0xA042U
 /** @} */
 
+/**
+ * @name    PIO instruction operands
+ * @note    The encoding is shared by several instructions but not every
+ *          value is legal in every one of them, see the datasheet
+ *          instruction set tables.
+ * @{
+ */
+#define PIO_DEST_PINS                   0U
+#define PIO_DEST_X                      1U
+#define PIO_DEST_Y                      2U
+#define PIO_DEST_NULL                   3U
+#define PIO_DEST_PINDIRS                4U
+#define PIO_DEST_PC                     5U
+#define PIO_DEST_ISR                    6U
+#define PIO_DEST_OSR                    7U
+
+/* EXEC is encoded differently by OUT and MOV.*/
+#define PIO_OUT_DEST_EXEC               7U
+#define PIO_MOV_DEST_EXEC               4U
+
+#define PIO_SRC_PINS                    0U
+#define PIO_SRC_X                       1U
+#define PIO_SRC_Y                       2U
+#define PIO_SRC_NULL                    3U
+#define PIO_SRC_STATUS                  5U
+#define PIO_SRC_ISR                     6U
+#define PIO_SRC_OSR                     7U
+
+#define PIO_MOV_NONE                    0U
+#define PIO_MOV_INVERT                  1U
+#define PIO_MOV_BITREV                  2U
+
+#define PIO_JMP_ALWAYS                  0U
+#define PIO_JMP_NOT_X                   1U
+#define PIO_JMP_X_DEC                   2U
+#define PIO_JMP_NOT_Y                   3U
+#define PIO_JMP_Y_DEC                   4U
+#define PIO_JMP_X_NE_Y                  5U
+#define PIO_JMP_PIN                     6U
+#define PIO_JMP_NOT_OSRE                7U
+
+#define PIO_WAIT_GPIO                   0U
+#define PIO_WAIT_PIN                    1U
+#define PIO_WAIT_IRQ                    2U
+#if defined(RP2350) || defined(__DOXYGEN__)
+#define PIO_WAIT_JMPPIN                 3U
+#endif
+
+/* ORed into the index of pioEncodeIrq()/pioEncodeWait() with
+   PIO_WAIT_IRQ: the flag index is taken relative to the state machine
+   number.*/
+#define PIO_IRQ_INDEX_REL               0x10U
+/** @} */
+
 /*===========================================================================*/
 /* Driver pre-compile time settings.                                         */
 /*===========================================================================*/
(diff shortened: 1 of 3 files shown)


chibios-sheriff (commented):

Scope: RP PIOv1 — a full set of PIO instruction encoders (`pioEncodeJmp/Wait/In/Out/Push/Pull/Mov/Irq/Set/Nop/Delay/SideSet`) plus the destination/source/condition constants. Base `master`.

> not the full review yet.

Cross-checked every encoder against the PIO instruction set — all correct:

- Opcodes and field placement: JMP `000` cond[7:5]/addr[4:0], WAIT `001` pol[7]/src[6:5]/index, IN/OUT `010`/`011` src|dest[7:5]/count[4:0], PUSH/PULL `100` with the `[7]` push/pull selector and IfFull/IfEmpty[6]/Block[5], MOV `101` dest[7:5]/op[4:3]/src[2:0], IRQ `110` clr[6]/wait[5], SET `111` dest[7:5]/data[4:0].
- The subtle bits are right too: `OUT` encodes EXEC as dest 7 while `MOV` uses dest 4 (correctly split into two constants), `PIO_IRQ_INDEX_REL = 0x10` (index bit 4), NOP = `mov y, y`, and the delay/side-set field is MSB-aligned in bits [12:8] with the bounds keyed off `sideset_bits`.
- Range checks (`cond < 8`, `addr < RP_PIO_NUM_INSTR_MEM`, delay/side-set fits the available bits) guard the debug builds.

Test 22 rebuilds the square-wave program entirely from the encoders and runs it, which is a good end-to-end check that the encodings match what the hardware executes.

emolitor:
Will review this evening.

emolitor:
Reviewed. Strongest of the four — this is the one I would merge first. Two documentation-level points, no functional defects found.

**Encodings verified against the instruction set tables.** I checked every encoder bit by bit:

- JMP `0x0000 | cond<<5 | addr`, WAIT `0x2000 | pol<<7 | src<<5 | index`, IN `0x4000`, OUT `0x6000`, PUSH `0x8000` / PULL `0x8080` with IfFull/IfEmpty at bit 6 and Block at bit 5, MOV `0xA000 | dest<<5 | op<<3 | src`, IRQ `0xC000 | clear<<6 | wait<<5 | index`, SET `0xE000` — all correct.
- `pioEncodeDelay()`: field at bits 12:8, range check `cycles < (1 << (5 - sideset_bits))` correct.
- `pioEncodeSideSet()`: correct in both directions — `value << (13 - sideset_bits)` puts the data at the top of the 5-bit field, and the SIDE_EN form sets bit 12 and shifts by `12 - sideset_bits`. Correct at both ends of the range (1 and 5 data bits).
- Both internal conversions are bit-identical: `pioSmSetPCX()` emits `0x0000 | addr` as before, and `pioSmSetConsecutivePindirsX()` emits `0xE000 | 4<<5 | value` = the old `0xE080 | value`.
- The operand tables match: `PIO_SRC_*` skips 4 for MOV (STATUS at 5), `PIO_MOV_*` rejects the reserved op 3, and splitting `PIO_OUT_DEST_EXEC` (7) from `PIO_MOV_DEST_EXEC` (4) is exactly right.

**1. `sideset_bits` means two different things in adjacent functions.** `pioEncodeDelay()` documents it as "enable bit included (0..5)"; `pioEncodeSideSet()` documents it as "enable bit excluded (1..5, 1..4 with `side_en`)". Both are individually correct — the delay field shrinks by the total side-set width while the side-set *value* width excludes the enable bit — but the same parameter name carrying opposite conventions in two functions a caller uses together on the same instruction is a trap. A caller with SIDE_EN and 3 data bits passes 4 to one and 3 to the other. Either rename one (`sideset_total` / `sideset_data`) or say so explicitly in both doc blocks.

**2. `pioEncodeMov()`'s destination doc is too loose.** It says "one of the `PIO_DEST_*` macros or `PIO_MOV_DEST_EXEC`", but for MOV the shared table does not apply: `PIO_DEST_PINDIRS` (4) is EXEC, and `PIO_DEST_NULL` (3) is reserved. A caller following the doc literally and passing `PIO_DEST_PINDIRS` to `pioEncodeMov()` gets `MOV EXEC` — which is about the most surprising possible outcome, and the `dest < 8U` check cannot catch it. Worth enumerating the legal MOV destinations the way `pioEncodeSet()` and `pioEncodeIn()` already enumerate theirs. This is the one place where the general "operand legality is the datasheet's domain" stance costs more than it saves, because the encoding actively collides with the table the doc points at.

**On your convention questions:** no X suffix is right — `PIO_INSTR_NOP` and `PIO_SM_CLKDIV()` are the precedent and these have no register effect. And folding `PIO_IRQ_INDEX_REL` into the index rather than taking a `bool rel` is the better call for the reason you give; it also leaves room for the RP2350 index modes without touching the signature.

Test 22 is the right design — checking the encoders against the literals the suite and the driver already use means the test would have caught a wrong encoder rather than encoding the same mistake twice.

emolitor (approved):

PR #232, opened 2026-08-12
Fix STM32U5 dynamic clock register handling

CInsights:
## Summary

  Correct STM32U5 clock-tree configuration writes that were being verified
  against complete register values even where the registers contain read-only or
  hardware-controlled status fields.

  The affected RCC and PWR writes now disable whole-register readback
  verification where the written value cannot be expected to match the complete
  register contents. Existing explicit readiness and clock-switch timeout checks
  remain in place and continue to verify the actual hardware transitions.

  `RCC_ICSCR1` is handled separately because it contains both configuration
  fields and factory-programmed MSI calibration fields. The change:

  - defines a mask containing only the writable MSI configuration fields;
  - updates those fields using `halRegModify32X()`;
  - preserves the factory-programmed `MSICALx` calibration values;
  - continues to verify the writable configuration fields;
  - retains the existing bounded waits for oscillator, regulator, PLL and
    system-clock transitions.

  This is needed because full-register verification can incorrectly fail when
  read-only status bits change asynchronously, and writing all of `ICSCR1` can
  overwrite factory calibration data.

  ## Related

  - Issue(s): None
  - Notes for reviewers:
    - The change is confined to the STM32U5 dynamic clock configuration path.
    - Readback verification is disabled only for registers containing
      hardware-controlled or read-only fields.
    - Explicit bounded readiness checks remain unchanged.
    - `ICSCR1` configuration fields are still verified through the masked
      modify operation.
    - Factory `MSICALx` fields are deliberately preserved.

  ## Target Branch Check

        merged on `master`

  ## Scope

  The implementation changes only
  `os/hal/ports/STM32/STM32U5xx/hal_lld.c`. Other STM32 ports are unaffected.

  ## Mechanical Checks

  The relevant STM32U5 ARM configurations compile successfully. The POSIX
  simulator was not rebuilt because this change is confined to the STM32U5 HAL
  clock code.

  ## Testing

  Tested through STM32U5 firmware builds and hardware startup. The STM32U5 clock
  configuration completes without the previous false register-verification halt,
  and the device proceeds through HAL initialization and normal operation.

  ## Compatibility and Risk

  No public interfaces or configuration structures are changed.

  The main risk is reduced immediate whole-register verification for the
  affected RCC and PWR writes. This is intentional because those registers
  contain status or read-only fields that make exact readback invalid. Critical
  state changes remain protected by the existing bounded waits for readiness and
  source-switch status.

  The `ICSCR1` write remains masked and verified because its writable
  configuration fields can be checked safely while preserving the factory
  calibration fields.

### DIFF
--- a/os/hal/ports/STM32/STM32U5xx/hal_lld.c
+++ b/os/hal/ports/STM32/STM32U5xx/hal_lld.c
@@ -64,6 +64,13 @@
 
 #define STM32_WS_THRESHOLDS             16U
 
+/* Writable configuration fields in RCC_ICSCR1. The MSICALx fields contain
+   factory calibration values and must be preserved.*/
+#define STM32_RCC_ICSCR1_CFG_MASK       (RCC_ICSCR1_MSIBIAS_Msk |          \
+                                         RCC_ICSCR1_MSIRGSEL_Msk |         \
+                                         RCC_ICSCR1_MSIKRANGE_Msk |        \
+                                         RCC_ICSCR1_MSISRANGE_Msk)
+
 #if (STM32_CLOCKTREE_VARIANT_U595 == TRUE) ||                               \
     (STM32_CLOCKTREE_VARIANT_U5A5 == TRUE)
 #define STM32_RCC_CFGR2_FIXED_BITS      0x00006000U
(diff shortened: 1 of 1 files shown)


gdisirio (changes requested):
The `ICSCR1` calibration preservation is the right direction, and the U575/U5A5 targets build cleanly. Please address the remaining STM32U5 points in this PR:

1. Mirror the clock-configuration changes into `os/xhal/ports/STM32/STM32U5xx/hal_lld.c`; its implementation is the classic HAL twin and retains the same affected writes.
2. Avoid disabling safety verification wholesale on `RCC_CR`, `RCC_CFGR1`, `PWR_VOSR`, and `RCC_BDCR`. Use writable-configuration masks with `halRegMaskedWrite32X()` or `halRegModify32X()` and keep `verify=true`, while retaining the existing readiness/status waits. In particular, `VOSRDY` can already be set if a requested VOS write does not latch, so that wait alone does not verify the selected voltage range.
3. Mask the requested `ICSCR1` value before using it as the modify set-mask, so bits outside `STM32_RCC_ICSCR1_CFG_MASK` cannot be written or included in verification.

STM32U3 can be audited separately; this review is limited to keeping the U5 fix complete and preserving its safety coverage.

gdisirio:
I prepared #235 as a concrete patch for the requested U5 changes. It keeps masked `verify=true` coverage for the affected RCC/PWR fields, masks the `ICSCR1` input, and mirrors the fix into XHAL. It passed style checks plus clean U575/U5A5 classic HAL and U575 XHAL builds, including `HAL_USE_SAFETY=TRUE` builds.

You can merge #235 into this branch or cherry-pick commit `7dba1d629b`.

gdisirio (approved):

PR #241, opened 2026-08-17
RT: establish initial SMP kernel lock ownership

gdisirio:
## Summary

- initialize the port before publishing an RT instance globally
- require the SMP initialization hook to establish startup lock ownership early
- make all three RP SMP back ends acquire the kernel hardware spinlock in `port_smp_init()`
- keep the simulator SMP port's first-unlock rendezvous as a valid alternate startup protocol

## Correctness rationale

RT instance initialization starts with the debug and statistics state representing an I-locked critical section, and its final `chSysUnlock()` releases the port lock. The merged RP SMP ports did not acquire that hardware spinlock during startup, so one core could release a lock held by the other while their instance initialization overlapped.

`chInstanceObjectInit()` now assigns the private core/configuration fields and calls `port_init()` before publishing the instance. The RP `port_smp_init()` implementations acquire the global kernel lock as their first operation, and the existing final `chSysUnlock()` is the matching release. Optional Cortex-M SMP hooks are also moved to the beginning of `port_init()`; ARMv7-R retains its prerequisite platform initialization first.

The simulator SMP branch does not use `port_smp_init()`. Its special first-unlock primary/secondary rendezvous already provides equivalent startup serialization, so it is intentionally unchanged.

## Validation

- `demos/RP/RT-RP2040-PICO`: ARMv6-M SMP build with optimization and LTO
- `demos/RP/RT-RP2350-PICO2`: ARMv8-M-ML-ALT SMP build with optimization and LTO
- `origin/chibios-smp-sim-dev/test/rt_smp`: full startup, kernel-lock stress, and IPI stress test with the generic instance-publication reorder applied
- `git diff --check origin/master...HEAD`

The Hazard3/RP2 path was source-checked; its cross compiler is not available in the current environment.

### DIFF
--- a/os/common/ports/ARMv6-M/smp/rp2/chcoresmp.c
+++ b/os/common/ports/ARMv6-M/smp/rp2/chcoresmp.c
@@ -104,13 +104,17 @@ static volatile uint32_t port_lockout_msg_count[PORT_CORES_NUMBER];
 /*===========================================================================*/
 
 static void port_local_halt(void) {
+  os_instance_t *oip;
   const char *reason = "remote panic";
 
   port_disable();
+  oip = currcore;
 
   __trace_halt("remote panic");
 
-  currcore->dbg.panic_msg = reason;
+  if (oip != NULL) {
+    oip->dbg.panic_msg = reason;
+  }
 
   CH_CFG_SYSTEM_HALT_HOOK(reason);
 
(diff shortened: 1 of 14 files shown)


emolitor (approved):
Overall the change looks good. One minor issue that is worth fixing since its a regression on the old code; In a debug build if I trace the life cycle of currcore in the old code it was null guarded, in the new code if an assert fires during port_init it will try to derefence a null currcore. I would null guard currcore ` if (currcore != NULL) currcore->dbg.panic_msg = reason; ` as well as in __trace_halt just to be defensive. Very minor though.

gdisirio:
Thanks, added guards for that.

PR #244, opened 2026-08-18
RT: remove redundant system-state publication hooks

gdisirio:
## Summary

- remove the optional `PORT_SYSTEM_STATE_RELEASE()` and `PORT_SYSTEM_STATE_ACQUIRE()` calls from RT startup
- remove the Hazard3 definitions which supplied the only merged implementation
- document that `chSysWaitSystemState()` observes the state value but does not itself enter a critical section

## Rationale

The hooks were introduced with the Hazard3 port in `861b12c3d6a`, before the initial SMP lock ownership correction in #241. At that time the state transition was being used as a publication handoff because instance startup did not acquire the shared kernel spinlock.

After #241, every merged secondary-core startup waits for `ch_sys_running` and immediately calls `chInstanceObjectInit()`. Its first port operation acquires the shared kernel spinlock. The primary stores `ch_sys_running` while owning that lock and then releases it through the final `chSysUnlock()`.

The spinlock release/acquire pair is therefore the authoritative synchronization point:

- it serializes primary and secondary instance startup
- its barriers publish the primary's preceding initialization before the secondary accesses shared kernel objects
- the state variable only tells the secondary when it may begin attempting instance initialization

Hazard3 already executes full fences in `port_spinlock_release()` and `port_spinlock_take()`. The separate state fences duplicate that ordering. ARM SMP startup follows the same lock handoff without defining these hooks.

@emolitor, please review in particular whether removing the original Hazard3 fences matches the intended post-#241 startup model.

## Validation

- `tools/style/stylecheck.py` on both changed files
- `demos/various/RT-Posix-Simulator`
- `demos/RP/RT-RP2040-PICO`
- `demos/RP/RT-RP2350-PICO2`

All build trees were cleaned after validation.

### DIFF
--- a/os/common/ports/RISCV-HAZARD3/chcore.h
+++ b/os/common/ports/RISCV-HAZARD3/chcore.h
@@ -250,21 +250,6 @@
 #define PORT_INFO                       "Compact kernel mode"
 /** @} */
 
-/**
- * @brief   Publishes initialization performed before a shared state flag.
- * @details The RP2350 cores are coherent, but RVWMO still requires an explicit
- *          release/acquire pair for message-passing through ordinary volatile
- *          storage.
- */
-#define PORT_SYSTEM_STATE_RELEASE()                                        \
-  __asm__ volatile ("fence rw, rw" : : : "memory")
-
-/**
- * @brief   Acquires initialization after observing a shared state flag.
- */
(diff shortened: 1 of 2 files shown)


emolitor (approved):
Yes, this is good and the hooks are not necessary.

PR #255, opened 2026-08-20
RP: fit HAZARD3-VALIDATION beside the core-local instance data

### LINKED ISSUE #254: HAZARD3-VALIDATION no longer links: core-local instances overflow the scratch banks
## Summary

`testhal/RP/RP2350/HAZARD3-VALIDATION` no longer links in its default configuration:

```
ld: section `.ram4_bss' will not fit in region `ram4'
ld: section `.ram5_bss' will not fit in region `ram5'
ld: region `ram4' overflowed by 1056 bytes
ld: region `ram5' overflowed by 1056 bytes
```

The failure is toolchain-independent: xPack `riscv-none-elf-gcc` 15.2.0 and a riscv-collab GCC 16.1.0 nightly produce the identical 1056-byte overflow. The tree at `385e972685` (where the validation suite landed) links cleanly with both compilers.

## Cause

`d81018df66` enabled the `CH_MEM_LOCAL_BSS()` / `CH_MEM_LOCAL_COHERENT_BSS()` annotations on the Hazard3 RP2 port, which places the per-core kernel instances (`ch0`/`ch1`, 1056 bytes each via `chsys.o` `.ram4_clear`/`.ram5_clear`) in the 4 KiB scratch banks. That commit re-aligned the startup defaults and `RT-RP2350-RISCV-BLINK` to the 0x400+0x400 stack split used by the ARM ports so that half of each bank stays free for annotated data — but `HAZARD3-VALIDATION`'s Makefile pins `USE_PROCESS_STACKSIZE` and `USE_EXCEPTIONS_STACKSIZE` at 0x800 locally. The two stacks fill the banks exactly and the instance data has no room.

The breakage went unnoticed because CI has no RISC-V toolchain leg.

## Fix

Align the validation Makefile to the same 0x400+0x400 split. Because the suite runs with `CH_DBG_ENABLE_STACK_CHECK=TRUE` and exercises the PMP stack guard, the smaller stacks need a hardware re-run of the full suite, not just a link check.

## Note

PR #251 adds panic regression modes to the same Makefile with their own scoped stack sizes; the changes are in different hunks and merge cleanly in either order.

emolitor:
Fixes #254

## Summary

- align the HAZARD3-VALIDATION per-core stacks to the 0x400+0x400 split established by `d81018df66` for the startup defaults and the RISC-V blink demo

Since `d81018df66` the core-local BSS annotations are live on the Hazard3 RP2 port, so the per-core kernel instances (1056 bytes each) share the 4 KiB scratch banks with the stacks. This Makefile still pinned 0x800+0x800, which fills the banks exactly and has made the default configuration fail to link ever since.

## Validation

- default configuration links with xPack riscv-none-elf-gcc 15.2.0 and with a riscv-collab GCC 16.1.0 nightly; both previously failed with identical 1056-byte `ram4`/`ram5` overflows
- full suite re-run on the Pico 2 bench in RISC-V mode with the reduced stacks; the configuration keeps `CH_DBG_ENABLE_STACK_CHECK=TRUE` and the PMP stack guard active, so undersized stacks would fault rather than pass silently
- no conflict with #251: that PR touches different hunks of this Makefile (regression-mode source selection); the branches merge cleanly in either order

### DIFF
--- a/testhal/RP/RP2350/HAZARD3-VALIDATION/Makefile
+++ b/testhal/RP/RP2350/HAZARD3-VALIDATION/Makefile
@@ -54,13 +54,13 @@ endif
 # Stack size to be allocated to the RISC-V process stack. This stack is
 # the stack used by the main() thread.
 ifeq ($(USE_PROCESS_STACKSIZE),)
-  USE_PROCESS_STACKSIZE = 0x800
+  USE_PROCESS_STACKSIZE = 0x400
 endif
 
 # Stack size to the allocated to the RISC-V main/exceptions stack. This
 # stack is used for processing interrupts and exceptions.
 ifeq ($(USE_EXCEPTIONS_STACKSIZE),)
-  USE_EXCEPTIONS_STACKSIZE = 0x800
+  USE_EXCEPTIONS_STACKSIZE = 0x400
 endif
 
 #


emolitor:
@gdisirio once this lands I will followup with another MR to add the riscv target to CI so this doesn't happen again.

gdisirio (approved):

PR #268, opened 2026-08-30
STM32U5: centralize OCTOSPI IRQ registration

CInsights:
## Summary

- Remove the legacy driver-conditioned OCTOSPI1/2 include, initialization, and deinitialization blocks from the STM32U5 IRQ dispatcher.
- Retain the unconditional platform-owned OCTOSPI registration added by PR #253.
- Let `stm32_octospi1.inc` and `stm32_octospi2.inc` decide whether each vector is emitted and enabled through their existing `STM32_OCTOSPIx_IS_USED` guards.

## Rationale

PR #253 added the current STM32 dispatcher pattern but left the older guarded registration in place, producing duplicate helper and vector definitions when WSPI was enabled. Removing the new unconditional path would fix the immediate compilation error but preserve the obsolete integration. The dispatcher should own the vector structure consistently; driver-use conditions belong inside the `.inc` handlers.

## Scope

- [x] No driver API or register configuration changes
- [x] Disabled and enabled OCTOSPI paths were checked

## Mechanical Checks

- [x] `git diff --check`
- [x] `tools/style/stylecheck.py os/hal/ports/STM32/STM32U5xx/stm32_isr.c`
- [x] STM32U575 HAL build with WSPI disabled
- [x] STM32U575 HAL build with WSPI/OCTOSPI1 enabled

The active-path ELF contains one strong `Vector170` for OCTOSPI1 while unused OCTOSPI2 remains the weak default `Vector220`.

## Testing Note

A minimal configuration with WSPI as the only DMA3 consumer exposes a separate pre-existing integration gap: `STM32U5xx/hal_lld.h` does not include `stm32_dma3.h` for `HAL_USE_WSPI` alone. The enabled-path build therefore also enabled SPI1 to use the existing DMA3 include path. That issue is independent of IRQ ownership and is intentionally not bundled here.

Bob's surrounding OCTOSPIv3 implementation has already passed STM32U5F7 NOR identification and sector read/write testing.

## Compatibility and Risk

No API or ABI changes are introduced. When an OCTOSPI instance is disabled, its `.inc` helper is a no-op and no strong handler is emitted. When enabled, the same handler and NVIC setup are emitted exactly once.


gdisirio:
@CInsights I revised this to retain the unconditional platform-owned OCTOSPI registration from #253 and remove the older driver-conditioned blocks instead. The `.inc` guards now own enabled/disabled selection, matching recent STM32 dispatchers. Disabled and OCTOSPI1-enabled builds pass; the enabled ELF has one strong `Vector170` and weak unused `Vector220`. Please review this revised direction.

gdisirio:
Added follow-up commit 0e0e021991 to centralize STM32_IRQ_OCTOSPI1_PRIORITY and STM32_IRQ_OCTOSPI2_PRIORITY under IRQ system settings. Both U575/U585 and U595/U5A5 mcuconf templates were updated, all four generated U5 mcuconf.h files were regenerated, and representative U575 and U5A5 builds pass.

gdisirio (approved):

PR #271, opened 2026-08-30
STM32: make grouped oscillator startup timeout configurable

CInsights:
## Summary

- Add `STM32_CFG_OSCILLATORS_STARTUP_TIME` to STM32 ports that start oscillators as a group: C0, G0, G4, H5, U3, and U5, in both HAL and XHAL.
- Default the setting to each family's existing `STM32_OSCILLATORS_STARTUP_TIME`, preserving current behavior unless a project overrides it.
- Keep all concurrently started U5 oscillators, including HSE, in one readiness wait and one deadline.
- Update the associated `mcuconf.h`/`xmcuconf.h` templates and regenerate all managed configuration files.

## Rationale

The group startup timeout represents the longest startup time among oscillators enabled together. A board with a slow HSE can now raise `STM32_CFG_OSCILLATORS_STARTUP_TIME` without introducing an HSE-only second wait or an additive deadline. For example:

```c
#define STM32_CFG_OSCILLATORS_STARTUP_TIME 20000U
```

## Scope

- [x] HAL and XHAL implementations are updated together
- [x] Configuration templates and generated files are updated together

The only incidental source cleanup removes stale commented-out debug lines from files touched by this PR so they pass the repository style checker.

## Mechanical Checks

- [x] `git diff --check`
- [x] `tools/style/stylecheck.py` on all 128 changed C/header files
- [x] Representative HAL builds: STM32C031, STM32G071, STM32G474, STM32H563, STM32U385, STM32U575
- [x] Representative XHAL builds: STM32C031, STM32G071, STM32G474, STM32H563, STM32U385, STM32U575

All build artifacts were cleaned after testing.

## Hardware Testing

On Bob's STM32U5F7 hardware with a 24 MHz crystal, debugger-assisted DWT measurements placed successful HSE startup near 9 ms. With a 20000U configured timeout and the DWT correction from PR #270, the setup passed 20 consecutive NRST boots.

## Compatibility and Risk

Existing configurations keep their previous family timeout because the new setting defaults to `STM32_OSCILLATORS_STARTUP_TIME`. Projects only need an override when their board oscillator startup characteristics exceed that default. The timeout remains a single shared deadline for the oscillator group, so fault-detection time is not accidentally compounded by sequential waits.


gdisirio:
@CInsights I generalized the HSE-specific change into `STM32_CFG_OSCILLATORS_STARTUP_TIME` for every STM32 HAL/XHAL port that starts oscillators as a group (C0/G0/G4/H5/U3/U5). U5 HSE is back in the common readiness mask, preserving a single group deadline; the family constant remains the default. Could you review this revised form, especially against your U5F7 hardware case?

gdisirio:
Follow-up `54da6497b1` makes the active generated clock models the source of truth: HAL U5 and XHAL G4/U3/U5 now declare `OSCILLATORS_STARTUP_TIME` in `<settings><configs>`, and their generated headers own the fallback. Manual fallbacks remain only in non-generated ports. XML validation, the clocktree suite, style checks, and all four representative builds pass.

CInsights:
Reviewed `4b7b0db303` and `54da6497b1`. The generalized `STM32_CFG_OSCILLATORS_STARTUP_TIME` setting is suitable for the custom U5 board case. We will set it to the board-validated `20000U`, replacing the local HSE-only override. Keeping HSE in the grouped readiness mask and making the generated clock model the source of the default are both preferable. The broad generated configuration update is expected.

gdisirio (approved):

PR #273, opened 2026-08-31
Add Cortex-R5 support to ARMv7-R port

arasnazmi:
## Summary

Adds Cortex-R5 support to the existing ARMv7-R port and an `ARMCR5` startup device
beside the existing `ARMCR8` and `RZV2H` devices.

The ARMv7-R port already supports the Cortex-R8. The R5 differs from the R8 only in
ways the startup and device layers handle, so the port itself needs just the variant
selection: an `ARM_CORE_CORTEX_R5` constant and its `PORT_CORE_VARIANT_NAME` branch,
six lines in `chcore.h`.

The startup side follows the existing `ARMCR8` device layout exactly. The Cortex-R5
has no VBAR, so the reset vector must live at address 0; `crparams.h` reflects that.
The shared `ARMCRx` startup files (`crt0_v7r.S`, `vectors.S`, `crt1.c`, the linker
rules and `arm-none-eabi.mk`) are **not touched** — the contribution is purely
additive.

Found and validated while bringing up a TI AM67A (J722S) Cortex-R5F on top of this
port; we are the second user of the ARMv7-R port after the R8 validation.

## Related

- Issue(s): none
- Notes for reviewers: this is the first of a stacked series adding a TI AM67 XHAL
  port. It has no dependencies and stands alone. The follow-up PRs build on it and
  should be merged in order.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

      all 3 changed `.c`/`.h` files: no findings
      `RT-Posix-Simulator` links; `test/rt-ports/testbuild` ARMv7-M target builds;
      `RT-ARMCR8-GENERIC` builds at **text 4900 B, byte-identical to the unpatched
      tree**, which is the evidence that the R5 work does not alter the existing R8
      path. Toolchain `arm-none-eabi-gcc 13.2.1` (Arm GNU Toolchain 13.2.rel1).
      Rebased onto and verified against `master` on 2026-08-30.

## Testing

- Any other tests run on a device? This PR adds no `main()` and no demo, so it
  produces no flashable image on its own. The ARMCR5 startup device it adds is
  exercised on real silicon by the demo added later in this series: on a T3 Gemstone
  O1 (AM67A/J722S) R5F, 2026-08-06, the RTOS heartbeat ran 96 iterations exactly
  1000 ms apart over 96 s with zero drift, which requires this startup device to have
  brought the core up correctly.

## Compatibility and Risk

Risk is low: every change is additive. No existing constant, macro or file changes
meaning, and the identical `RT-ARMCR8-GENERIC` binary size demonstrates that existing
configurations generate the same code as before.

### DIFF
--- a/os/common/ports/ARMv7-R/chcore.h
+++ b/os/common/ports/ARMv7-R/chcore.h
@@ -85,6 +85,7 @@
  * @name    ARMv7-R variants
  * @{
  */
+#define ARM_CORE_CORTEX_R5              5
 #define ARM_CORE_CORTEX_R8              8
 /** @} */
 
@@ -102,6 +103,11 @@
   #define PORT_ARCHITECTURE_ARM_CORTEXR8
   #define PORT_ARCHITECTURE_NAME        "ARMv7-R"
   #define PORT_CORE_VARIANT_NAME        "ARM Cortex-R8"
+
+#elif CORTEX_MODEL == ARM_CORE_CORTEX_R5
+  #define PORT_ARCHITECTURE_ARM_CORTEXR5
+  #define PORT_ARCHITECTURE_NAME        "ARMv7-R"
+  #define PORT_CORE_VARIANT_NAME        "ARM Cortex-R5"
 
 #else
   #error "unknown or unsupported ARMv7-R core"
--- a/os/common/startup/ARMCRx/compilers/GCC/mk/startup_armcr5.mk
+++ b/os/common/startup/ARMCRx/compilers/GCC/mk/startup_armcr5.mk
@@ -1,0 +1,17 @@
+# List of the ChibiOS generic Cortex-R5 startup files.
+STARTUPSRC = $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC/crt1.c
+
+STARTUPASM = $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC/crt0_v7r.S \
+             $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC/vectors.S
+
+STARTUPINC = $(CHIBIOS)/os/common/portability/GCC \
+             $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC \
+             $(CHIBIOS)/os/common/startup/ARMCRx/devices/ARMCR5 \
+             $(CHIBIOS)/os/common/ext/ARM/CMSIS6/Include
+
+STARTUPLD  = $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC/ld
+
+# Shared variables
+ALLXASMSRC += $(STARTUPASM)
+ALLCSRC    += $(STARTUPSRC)
+ALLINC     += $(STARTUPINC)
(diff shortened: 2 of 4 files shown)


gdisirio (approved):

gdisirio:
I found two issues in the Cortex-R5 startup path:

- crt0_v7r.S initializes VBAR by default, but Cortex-R5 has no VBAR. This would trigger an Undefined Instruction exception during startup unless explicitly disabled.
- Both crt0_v7r.S and chcoreasm.S still declare .cpu cortex-r8, so R5 builds are assembled and identified as Cortex-R8.

I can prepare a fix for both issues if you would like.

arasnazmi:
Yes, we’d really appreciate your help with this. If you’re willing to prepare the fixes for both issues, we’d be very happy to review and test them on our side.

gdisirio:
A clarification/correction to my earlier review: the .cpu cortex-r8 directives in crt0_v7r.S and chcoreasm.S are intentional and should remain. Those modules are shared by the Cortex-R variants, and the assembler architecture is selected to admit the superset of instructions that the shared sources may use; it is not meant to identify the concrete runtime core.

Core differences must instead be handled inside the shared modules through conditional assembly. The actual startup issue remains the unguarded VBAR initialization: crt0_v7r.S defaults CRT0_VBAR_INIT to TRUE, and the new ARMCR5 crparams.h currently does not communicate a VBAR capability to that source. R5 support should therefore add an internal core/capability conditional in the shared startup so the VBAR instructions are omitted automatically for R5, rather than changing .cpu or requiring every R5 application to provide an override.

This means the contribution cannot be entirely additive if the shared startup needs adaptation, but the R8-generated path can remain unchanged. I will prepare this fix.

gdisirio:
I have prepared and validated the fixes for the shared Cortex-R startup and capability handling. They make VBAR initialization capability-driven, keep the assembler architecture selection as the shared superset, and make the device parameter headers expose only fixed CORTEX_* capabilities.

Because “Allow edits from maintainers” is currently disabled, there are two ways to update this PR:

1. Enable maintainer edits on the PR. I will cherry-pick the fixes onto the existing PR head and push the update directly.
2. Leave maintainer edits disabled. I will publish the two fix commits on a chibios-upstream branch and provide the exact cherry-pick commands so you can update the PR branch yourself.

Please choose whichever option you prefer; I will take care of the corresponding preparation.

arasnazmi:
One clarification: "Allow edits from maintainers" isn't something we disabled. The head branch is in `t3gemstone/chibios`, an organization-owned fork, and GitHub doesn't offer that checkbox for org forks at all.

I've sent you a collaborator invitation with **Write** access on the fork instead, same effect. Once you accept, cherry-pick onto `pr2-armv7r-cortex-r5` and push directly and the PR will update in place. If you'd rather not take repo access, option 2 works for us too.

We'll re-validate on our AM67A R5F once the commits land. Same applies to the follow-up PRs in the series.

gdisirio:
Thanks for the clarification; that explains why the maintainer-edit option is unavailable. There is no need to grant access to the organization fork. I have restored the Cortex-R5 series and the two shared-startup corrections on top of current master. We will keep this contribution intact and handle the VBAR capability guard and immutable device capabilities as a focused maintainer follow-up, then ask you to revalidate the combined result on the AM67A R5F. The collaborator invitation can therefore be revoked.

gdisirio:
The follow-up has now been merged immediately after this PR as #285. Current master contains the Cortex-R5 contribution at 6e4c21cfdb and the shared-startup corrections at 52dbe22d9c. @arasnazmi, when convenient, please revalidate current master on the AM67A R5F and report the result here. No branch update or cherry-pick is needed on your side.

PR #233, opened 2026-08-15
Add LPTIM systick (initially for U5)

CInsights:
## Summary

Adopts the reduced-surface STM32 SYSTICKv3 LPTIM implementation proposed by
Giovanni in `chibios-lptim-systick-dev` at `6c312b8e1d`.

- SYSTICKv1 remains the default STM32U5 system-timer backend.
- `USE_ST_LPTIM=yes` selects SYSTICKv3 explicitly.
- The existing `STM32_ST_USE_TIMER` setting selects LPTIM1, LPTIM3 or LPTIM4;
  LPTIM2 is rejected because it is not in the Smart Run Domain.
- The prescaler is derived from the selected LPTIM clock and
  `OSAL_ST_FREQUENCY`, with unsupported ratios rejected at compile time.
- The LPTIM runs continuously as a 16-bit free-running system timer and CCR1
  provides the single ChibiOS alarm.
- Timestamp maintenance and platform wake policy remain application concerns.

Compared with `6c312b8e1d`, this branch contains two functional adjustments so
an application can independently use other LPTIM interrupt sources such as
CCR2:

1. CCR1 alarm start/stop operations preserve every non-`CC1IE` bit in `DIER`.
2. The SYSTICKv3 ISR dispatches `CC1IF` only when `CC1IE` is also enabled.

The first change prevents a ChibiOS alarm operation from disabling an
application-owned channel. The second prevents a stale disabled CCR1 flag from
being dispatched when another enabled source shares the LPTIM vector.

## Hardware feedback

The reduced driver plus these two adjustments is running on a custom STM32U575
board with:

- LPTIM3 clocked from LSE / 32 at 1024 Hz;
- 16-bit free-running mode;
- `CH_CFG_ST_TIMEDELTA=8` (the proposed conservative minimum);
- ChibiOS alarms on CCR1; and
- application-owned half-range timestamp maintenance on CCR2, serviced from
  `CH_CFG_IRQ_PROLOGUE_HOOK()`.

An ongoing automated test repeatedly schedules a 40,000 ms CCR1 alarm, enters
STOP2, handles the intervening CCR2 timestamp wake without restoring the RUN
clock tree, and then restores the clock tree for the CCR1 wake. At the latest
sample it had completed 219/219 iterations, 438 successful STOP2 clock-restore
attempts and 279 CCR2 maintenance interrupts with zero phase recoveries, clock
restore failures, UsageFaults, HardFaults or retained halts. No alarm-update
stall has occurred at the eight-tick minimum.

[...]

PR #274, opened 2026-08-31
demos: add STM32U575 SYSTICKv3 SRD example

CInsights:
## Summary

Adds a standalone `RT-STM32U575ZI-NUCLEO144-SRD` demo for the reduced
SYSTICKv3 LPTIM backend in #233.

- Selects SYSTICKv3 with `USE_ST_LPTIM=yes` and uses LPTIM3 from LSE / 32 as a
  1024 Hz, 16-bit ChibiOS system timer.
- Leaves CCR1 under the formal SYSTICKv3 driver for normal ChibiOS alarms.
- Implements half-range timestamp maintenance on application-owned CCR2 from
  `CH_CFG_IRQ_PROLOGUE_HOOK()`, outside the formal driver.
- Demonstrates timed STOP2 entry from the idle hooks and wake through a virtual
  timer.
- Uses the same common IRQ-prologue hook as the application STOP2 wake policy
  for all normal/kernel-aware IRQs.
- Keeps priority-zero fast handlers outside that policy.
- Includes an optional LPTIM1-paced circular LPDMA/LPGPIO activity probe on
  PA1 which continues while the core is in STOP2.

## Application-owned CCR2

The demo schedules CCR2 at `TIME_MAX_SYSTIME / 2`, calls
`chVTGetTimeStampI()` under a balanced ISR lock, and derives each next compare
from the previous compare to remain phase-locked. If servicing is late enough
to make the next compare invalid, it reschedules relative to the current
counter and reports a recovery count.

The common IRQ hook identifies the active LPTIM3 vector and qualifies both
compare flags with their corresponding enable bits. A CCR2-only STOP2 wake
performs timestamp maintenance without restoring the RUN clock tree. A CCR1
wake, including a simultaneous enabled CCR1+CCR2 case, restores the default RUN
clock tree before the SYSTICKv3 driver dispatches the virtual-timer callback.

This demonstrates the intended separation in #233: the driver owns CCR1 but
preserves non-CCR1 `DIER` bits, while the application can own CCR2 without the
driver enabling, disabling or dispatching it.

## STOP2 wake policy

`CH_CFG_IRQ_PROLOGUE_HOOK()` runs after ChibiOS statistics/trace/debug entry and
before the IRQ driver body. Clock restoration uses PRIMASK because a ChibiOS
BASEPRI kernel lock would not mask priority-zero autonomous interrupts during
RCC and Flash reconfiguration. Fast priority-zero handlers bypass the common
prologue and remain outside this wake policy.

The LPTIM1/LPDMA/LPGPIO probe alternately sets and clears LPGPIO1_P0. PA1 is
configured as AF11 and exposes a 1 Hz waveform for scope observation. The DMA
descriptor and source words are in SRAM4.

## Related

- Depends on #233 and targets `feature-U5-SYSTICKv3-LPTIM` so this stacked
  review contains only demo changes.
- Retarget to `master` after #233 is merged.
- The onboard green LED is a post-wake indication; PA1 is the autonomous
  no-wake probe.

## Target Branch Check

- [ ] This PR temporarily targets the #233 feature branch and must be retargeted
  to `master` before merge.

## Scope

- [x] I reviewed impact on other ports/configurations where relevant.

## Mechanical checks

- [x] `git diff --check` passes for the demo delta.
- [x] A clean SYSTICKv3/LPTIM3 build passed with ARM GCC 15.2.1 and xPack
  build-tools 4.4.1-2.1.

## Hardware evidence

The same driver/application split is under continuing automated test on a
custom STM32U575 board at `CH_CFG_ST_TIMEDELTA=8`. Each iteration schedules a
40,000 ms CCR1 alarm, enters STOP2, handles an intervening CCR2-only timestamp
wake, and then handles the CCR1 wake. At the latest sample, 219/219 iterations,
438 clock restores and 279 CCR2 interrupts had passed with no phase recovery,
clock failure or ARM fault.

Exact execution of this standalone demo and its PA1 probe on a
NUCLEO-U575ZI-Q remains pending.

## Compatibility and limitations

The demo depends on #233. Application CCR2 ownership is deliberate demo code,
not part of the SYSTICKv3 public API. The wrapping 16-bit system-time counter is
not used as an application clock; elapsed time uses `systimestamp_t`,
`chVTGetTimeStamp()` and `chTimeStampDiffX()`.


CInsights:
Updated this stacked branch to the current #233 head and current `master` ancestry.

The demo now checks `ST_LLD_REQUIRES_APPLICATION_TIMESTAMP` at compile time and uses `chVTSetContinuous(..., TIME_MAX_SYSTIME / 2, ...)` for the mandatory LPTIM4 timestamp updater. Its callback still calls `chVTGetTimeStampI()` in I-class context. The copied `chconf.h` also now supplies current master’s required `CH_CFG_USE_RFCU` option.

Targeted style and whitespace checks pass. Clean ARM GCC 15.2.1 builds pass for the committed LPTIM4/application-VT configuration and for a temporary LPTIM3 configuration exercising #233's automatic CCR2 path. Equivalent LPTIM4/LSE/32 behavior has passed a 70-second cross-wrap STOP2 test on a custom STM32U5 board; testing of this exact demo and PA1 probe on the NUCLEO remains pending.

CInsights:
Updated this stacked branch for the revised normal-IRQ wake policy and current #233/`master`.

- Replaced the demo's timer-only `STM32_ST_LPTIM_WAKEUP_HOOK()` wiring with `CH_CFG_IRQ_PROLOGUE_HOOK()` so every normal/kernel-aware IRQ restores RUN clocks before its driver body.
- Preserved the lightweight path for a two-channel LPTIM CCR2-only timestamp wake; simultaneous CCR1+CCR2 takes the restore path.
- Kept priority-zero fast handlers outside the ChibiOS prologue.
- Synchronized the copied `chconf.h` with the post-#293/#294 hook contracts and corrected the documented ordering.

Style and whitespace checks pass. Clean ARM GCC 15.2.1 builds passed for the committed LPTIM4/application-VT configuration and a disposable LPTIM3/automatic-CCR2 configuration. Disassembly confirms the common hook is reached by both the LPTIM system-timer and USART1 vectors before their driver bodies. All GitHub checks are green.

The PR description now records the existing custom-board RTC, non-timer and mixed-wake validation. Exact NUCLEO-U575ZI-Q execution and the simultaneous CCR1+CCR2 device case remain pending.

PR #279, opened 2026-09-01
hal: add portable WSPI status polling

CInsights:
# Summary

Integrate Giovanni's revised portable WSPI status-polling implementation from
`chibios-xhal-dev` through commit `0216b21273f9`.

The implementation was developed first in XHAL and then ported to the classic
HAL with matching semantics. It provides a blocking byte-vector operation that
polls until `(status & mask) == match`. Both the polling interval and timeout
use ChibiOS interval types.

Drivers without acceleration use ordinary receive operations and optional
thread sleeps. STM32 OCTOSPIv2 and OCTOSPIv3 accelerate supported one-to-four
byte operations in hardware and convert the portable interval to peripheral
clock cycles internally.

There is no public cancellation API. For an accelerated operation, a status
match or hardware error wakes the suspended thread. If the supplied timeout
expires, the same thread wakes with `MSG_TIMEOUT`. In all cases that waiter is
the sole owner of `wspi_lld_stop_status_poll()` and the final driver-state
restoration, avoiding the former double-abort and early-reuse races.

The XHAL implementation is generated from `os/xhal/codegen/hal_wspi.xml`; the
corresponding generated header and source are included. Classic HAL and XHAL
also have equivalent OCTOSPIv2/OCTOSPIv3 acceleration and generic fallbacks.

The N25Q and MX25 XSNOR Page Program paths consume the portable operation in
both HALs. Their SPI loops and asynchronous erase/query behavior are unchanged,
and device-specific error/security-register checks still run after the ready
match. The upstream consumers currently pass `TIME_INFINITE`.

This replaces the implementation in the first two PR commits. The current diff
against `master` is Giovanni's 24-file WSPI/XSNOR change set; the obsolete
external-cancellation API is not present.

## Related

- Builds on the OCTOSPIv3 configuration support merged in #253 and the
  centralized STM32U5 OCTOSPI IRQ registration merged in #268.
- Project-only realtime instrumentation is intentionally excluded.

## Target Branch Check

- [x] This PR targets `master`.
- [x] The PR branch includes current `master` through `96e51d2990`.

## Mechanical and Build Checks

- [x] `git diff --check origin/master..HEAD` passes.
- [x] The final WSPI/XSNOR files match Giovanni's `chibios-xhal-dev` branch at
  `0216b21273f9`.
- [x] Classic HAL STM32H735 WSPI-MFS/XSNOR builds and links with OCTOSPIv2,
  including the N25Q and MX25 consumers.
- [x] The custom board's `hw_test`, `runtime`, and `master` configurations build
  after porting the classic-HAL WSPI and OCTOSPIv3 implementation.
- [x] GitHub style, ARM smoke, and POSIX simulator checks pass.

## Device Testing

The revised classic-HAL/OCTOSPIv3 implementation was tested on a custom STM32U5
board with two interleaved W25Q256JV devices. The four classic-HAL/OCTOSPIv3
files from Giovanni's `0216b21273f9` branch were first layered unchanged onto
the custom board's ChibiOS snapshot.

- With `TIME_INFINITE`, a 48,732-byte JPEG programmed in 43 ms at
  1,134,126 bytes/s and passed full byte and CRC verification.
- With a finite five-millisecond kernel timeout, a 52,185-byte JPEG programmed
  in 962 ms at 54,251 bytes/s and passed full byte and CRC verification.
- A test-only one-shot non-matching status expectation forced the first Page
  Program through `MSG_TIMEOUT`. The store failed cleanly.
- Without resetting the board, the immediate retry programmed 51,917 bytes in
  962 ms and passed byte and CRC verification. This verifies timeout cleanup
  and driver reuse under the waiter-owned model.

The large finite-timeout performance cost is specific to repeatedly programming
the custom board's 1024 Hz system-timer alarm; it is not an OCTOSPI matcher
regression. The normal/infinite path retains the expected hardware-polling
throughput.

## Compatibility and Risk

- The API introduced by the first draft was replaced before merge; no released
  API is changed.
- `WSPI_LLD_SUPPORTS_STATUS_POLL` is explicitly defined by each LLD. Ports
  without acceleration receive the generic implementation.
- Accelerated polling is limited internally to the status length and interval
  supported by the selected OCTOSPI peripheral.
- The waiter owns cleanup after match, error, or timeout; no external path can
  publish `WSPI_READY` while the original thread is still suspended.


gdisirio:
One additional API-design concern: automatic masked status matching appears to be an STM32 controller feature rather than a common xSPI controller capability. I checked representative TI OSPI, Infineon SMIF, and Renesas OSPI_B implementations; their official SDKs implement readiness polling by repeatedly issuing ordinary status reads in software. As currently shaped, especially the 1..4-byte restriction and the interval expressed in peripheral-clock cycles, the high-level WSPI API exposes STM32-specific details and leaves other LLDs without a natural implementation.

The operation itself is portable: poll a status value until (status & mask) == match. I think we could keep that semantic operation in the HAL, guarantee an implementation through a generic software fallback using normal command/read transfers, and let capable LLDs such as STM32 OCTOSPIv3 accelerate it in hardware. The public interval should then use a portable time unit, with the LLD converting it to controller clocks and applying hardware limits internally. An asynchronous fallback would need a timer/state machine; alternatively, the HAL could provide a portable blocking poll operation and reserve asynchronous matching as an optional accelerated extension.

If you agree with this direction, we can address it in this PR; I am happy to help adapt the API and implementation.

gdisirio:
There is also a merge-blocking correctness issue in the cancellation path. wspiAbortStatusMatchI() aborts the peripheral, sets the driver to WSPI_READY, and resumes the waiting thread with MSG_TIMEOUT. When wspiStatusMatchTimeout() wakes with MSG_TIMEOUT, it unconditionally calls wspi_lld_abort_status_match() again. A single external cancellation therefore executes the low-level abort/restore sequence twice.

More importantly, WSPI_READY is published before the original waiter has finished. If the driver is reused before that waiter runs, the old waiter can execute its second abort against the subsequent operation and restore the saved status-match CR over its configuration.

The cleanup needs a single owner, or the waiter needs to distinguish its own timeout from an external abort that has already stopped the hardware. For example, the external abort path could stop the peripheral and record/resume with a distinct internal cancellation state, while only a timeout generated by osalThreadSuspendTimeoutS performs the low-level abort. The public result can still be MSG_TIMEOUT if that is the desired API contract.

We should address this along with the portable software-fallback design; I am happy to help with the change.

gdisirio:
One more scope point: XHAL is also within our perimeter, so this should not evolve only in the classic HAL API. Once we settle the portable status-polling semantics and cancellation model, the corresponding XHAL WSPI API/code generator should be updated as well, including the generic software fallback and hardware-assisted LLD implementation where supported. Keeping both HALs aligned now will avoid introducing two incompatible WSPI models. We can include the XHAL side in the work discussed above.

gdisirio:
XSNOR should be included as a consumer of the portable API as well. Both the legacy HAL and XHAL N25Q/MX25 implementations currently perform software polling after each programmed page: N25Q waits for the program/erase-ready flag, while MX25 waits for WIP to clear. On the WSPI path those loops can use the new status-polling operation, gaining hardware acceleration where available and the HAL software fallback everywhere else. The SPI path can retain its existing loop, and the device-specific error/security-register checks still need to run after the ready match.

Erase should remain as it is. It is intentionally modeled as a potentially very slow asynchronous operation: query_erase() performs one probe and returns FLASH_BUSY_ERASING, and flashWaitErase() sleeps between queries. That also releases a shared bus between probes rather than monopolizing it for the full erase duration.

We can include these XSNOR updates for both HAL and XHAL when addressing the API.

CInsights:
Addressed the feedback in `6845deb65f`:

- Replaced the STM32-shaped API with portable blocking byte-vector polling,
  `(status & mask) == match`, an arbitrary public length, and a microsecond
  interval.
- Added an ordinary-receive/thread-sleep fallback. OCTOSPIv3 acceleration is
  now optional and applies its 1..4-byte/PIR limits internally.
- Fixed cancellation ownership: the external I-class abort can stop and wake
  an accelerated poll, but does not publish READY; the original waiter alone
  performs final cleanup. The LLD-active flag also prevents double abort.
- Added the equivalent XHAL API to the generator, regenerated the checked-in
  files, and added XHAL OCTOSPIv3 acceleration.
- Migrated the N25Q and MX25 WSPI page-program ready loops in both HALs. SPI
  loops and erase/query behavior remain unchanged, and error/security checks
  still run after matching.

Local validation passed for XML schema validation, repeatable FMPP generation,
classic/XHAL STM32H735 WSPI-XSNOR integration builds (generic fallback), and
classic STM32U5 plus XHAL STM32H563 OCTOSPIv3 object builds. The earlier device
results cover the original hardware mechanism; I have called out in the PR
description that the portable revision still needs a device rerun.

gdisirio:
This needs more work, I am moving this idea into the chibios-xhal-dev branch. There are a couple of race conditions to be addressed and removing the cancellation API would make things simpler. I will proceed with copying incrementally parts of this PR into that branch.

gdisirio:
My fixes are in branch chibios-xhal-dev, give it a try when you have time.

CInsights:
Thanks Giovanni. I have updated this PR branch to incorporate your exact
`chibios-xhal-dev` WSPI/XSNOR series through `0216b21273f9`, merged with current
`master`. The resulting PR diff is your 24-file change set; the old public
cancellation API and its ownership model are gone.

Although the custom board does not use XHAL, it provided a hardware test of the
corresponding classic-HAL port on STM32U5 OCTOSPIv3 with two interleaved
W25Q256JV devices:

- `TIME_INFINITE`: 48,732 bytes programmed in 43 ms, full byte/CRC verification
  passed.
- Finite five-millisecond timeout: 52,185 bytes programmed in 962 ms, full
  byte/CRC verification passed.
- A forced one-shot non-match returned through the timeout path and failed the
  first store cleanly. An immediate retry without reset programmed 51,917 bytes
  and passed byte/CRC verification, confirming waiter-owned cleanup and driver
  reuse.

The current branch also passes a classic-HAL STM32H735 OCTOSPIv2
WSPI-MFS/XSNOR build covering the N25Q and MX25 consumers.

gdisirio (approved):

PR #289, opened 2026-09-04
T3_GEM_O1: add the XHAL port for the Cortex-R5F, the board and the demo

arasnazmi:
## Summary

Adds the foundation of a TI AM67 (J722S) XHAL port for the SoC's Cortex-R5F, plus the
board and a demo that exercises it. This is the first TI port in XHAL; it follows the
layout the RP port established — reusable IP blocks under
`os/xhal/ports/TI/LLD/<IPvN>/`, family glue under `os/xhal/ports/TI/AM67/`.

What lands here:

- **ARMv7-R AM67 sub-port** (`os/common/ports/ARMv7-R/platforms/am67/`) on the port's
  existing `port_platform_init()` hook, so VIM (Vectored Interrupt Manager) setup and
  `__port_irq_dispatch()` live in the sub-port rather than in the HAL.
- **XHAL platform** (`os/xhal/ports/TI/AM67/`): `hal_lld`, the instance registry and
  `platform.mk`.
- **DMTIMERv1 ST driver**, the system tick.
- **UARTv1 SIO driver**, the console path.
- **Board `T3_GEM_O1`** and the demo `RT-XHAL-T3-GEM-O1`, a heartbeat
  thread plus a console echo loop over `SIOD1`.

Three UART details shape the driver, since the AM67 UARTs are 16550-compatible with TI
extensions. `IER` shares an address with the divisor latch high byte selected by
`LCR.DLAB`, so the driver keeps a shadow instead of reading the register back. `LSR`
clears its error bits on read, so every read goes through a helper that latches them
for later reporting. And the 16550 has no line-idle status bit, so the RX idle event
is derived from the character timeout interrupt and recorded in a driver-private
sticky bit above the eight hardware `LSR` bits. The VIM line is level sensitive and no
FIFO condition clears until the application moves data, so the handler masks the
sources it serviced and the read/write paths re-arm them; without that the handler
re-enters until the application happens to drain the FIFO. `MDR1` is written last on
start and first on stop, because the peripheral is inert until the mode is selected.

The board file carries the peripheral input clocks the XHAL drivers ask for. They are
board-level rather than SoC-level because the device tree owns the clock muxes, so the
same peripheral on another carrier can be parented differently. The eHRPWM and eCAP
counter clocks are not among them: those are owned by `am67_registry.h`.

The demo's linker script gives code and data separate memory regions in both TCM and
DDR, so the image links as read-execute and read-write segments rather than one RWX
`PT_LOAD`. The MPU covers DDR with a single region and the split stays inside the same
remoteproc carveout, so nothing depends on the boundary.

## Related

- Issue(s): none
- Notes for reviewers: stacked on the AM67R5F startup device PR and should be merged
  after it. All files added here are new; no existing port, driver or demo is
  modified.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

      all 19 changed `.c`/`.h` files: no findings
      `RT-Posix-Simulator` links, `test/rt-ports/testbuild` ARMv7-M target builds, and
      `RT-XHAL-GEMSTONE-O1-R5F` builds warning-free at text 11544 B with
      `arm-none-eabi-gcc 13.2.1`

## Testing

- Any other tests run on a device? Yes, on a T3 Gemstone O1 (AM67A/J722S) R5F:
  - The demo ran with the heartbeat printing iterations 0 through 96 exactly 1000 ms
    apart across 96 s with zero drift, and later for 157 s unbroken with a clean fault
    block. That exercises ARMCR5 startup, the ARMv7-R AM67 sub-port and its
    `__port_irq_dispatch()`, the DMTIMER tick rate (confirming `AM67_ST_TIMER_CLOCK`
    = 25 MHz), XHAL core with `drvStart()`, and the scheduler.
  - The UART was verified separately, because a running heartbeat does not prove the
    UART transmits: the thread blocks until the TX FIFO accepts every byte, but a dead
    peripheral can leave `SSR.TXFIFOFULL` reading 0 forever and the driver then writes
    into a void without stalling. The UART's own internal loopback (`MCR.LPBK`) was
    used instead: wrote 4 bytes, read back 4, `'PING'`, `LSR = 0x60` (`THRE|TEMT`, no
    error bits). That covers TX, RX, both FIFOs and the driver's read and write paths.

## Compatibility and Risk

Everything added here is new files under new paths, so the only contact with existing
code is the ARMv7-R platforms hook, used as designed.

Bytes on the wire and the RX interrupt path are not proven. The loopback test is
conclusive for the driver's own paths but its reads were polled, and it does not put a
signal on a pin. A cable test is not possible on this board as configured: the console
TX pad is assigned to EHRPWM0 channel B by the device tree overlay, so no UART pin is
exposed. `MCR.LPBK` is retained in `ti_uart.h` because it is the cable-free way to
retest this driver.

### DIFF
--- a/demos/various/RT-XHAL-T3-GEM-O1/AM67_R5F.ld
+++ b/demos/various/RT-XHAL-T3-GEM-O1/AM67_R5F.ld
@@ -1,0 +1,152 @@
+/*
+    ChibiOS - Copyright (C) 2006-2026 Giovanni Di Sirio.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/*
+ * T3 Gemstone O1 (TI AM67A/J722S) Cortex-R5F memory setup.
+ *
+ * The image is loaded by the Linux k3-r5 remoteproc driver, addresses are
+ * device addresses as seen by the R5F core and must match the reserved
+ * memory carveouts in the host device tree.
+ */
+
+__und_stack_size__ = DEFINED(__und_stack_size__) ? __und_stack_size__ : 0x100;
+__abt_stack_size__ = DEFINED(__abt_stack_size__) ? __abt_stack_size__ : 0x100;
+__fiq_stack_size__ = DEFINED(__fiq_stack_size__) ? __fiq_stack_size__ : 0x100;
+__irq_stack_size__ = DEFINED(__irq_stack_size__) ? __irq_stack_size__ : 0x400;
+__svc_stack_size__ = DEFINED(__svc_stack_size__) ? __svc_stack_size__ : 0x100;
+__sys_stack_size__ = DEFINED(__sys_stack_size__) ? __sys_stack_size__ : 0x800;
+
+MEMORY
+{
+    /* The MCU domain core has a single usable TCM, mapped at address 0 by
+       ti,loczrama = <0> (the ATCM lands at 0x41010000 and is left disabled
+       by ti,atcm-enable = <0>). Vectors, the pre-MPU boot code and the
+       banked mode stacks therefore share it.
+
+       It is split into two regions so that the vectors are guaranteed to
+       land at address 0, where the Cortex-R5 fetches its reset vector (it
+       has no VBAR). A single region would not do: the sections below are
+       allocated in script order, so .tcm_probe would take address 0 and
+       push the vectors out of reach.
+
+       The top 64 bytes are left out of both regions, reserved for the boot
+       witness and fault blocks written by board.c and for the temporary
+       boot stack that grows down from there.*/
+    tcm_vec   (rx)  : org = 0x00000000, len = 256
+    tcm_code  (rx)  : org = 0x00000100, len = 1k
+    tcm       (rw)  : org = 0x00000500, len = 32k - 256 - 1k - 64
+    ddr_rsc   (rw)  : org = 0xA1100000, len = 4k    /* resource table      */
+    ddr_trace (rw)  : org = 0xA1110000, len = 16k   /* remoteproc trace    */
+    /* Code and data get separate regions so that the linker can emit a
+       read-execute segment and a read-write one instead of merging both
+       into a single RWX PT_LOAD. Nothing else depends on the boundary:
+       the MPU covers the whole of DDR with one region, and the split
+       stays inside the same remoteproc carveout.*/
+    ddr_code  (rx)  : org = 0xA1240000, len = 256k   /* text, rodata       */
+    ddr_data  (rw)  : org = 0xA1280000, len = 13824k /* data, bss, heap    */
+
+    /* Unused regions referenced by the generic memory rules.*/
+    flash0 (rx) : org = 0x00000000, len = 0
+    flash1 (rx) : org = 0x00000000, len = 0
+    flash2 (rx) : org = 0x00000000, len = 0
+    flash3 (rx) : org = 0x00000000, len = 0
+    flash4 (rx) : org = 0x00000000, len = 0
+    flash5 (rx) : org = 0x00000000, len = 0
+    flash6 (rx) : org = 0x00000000, len = 0
+    flash7 (rx) : org = 0x00000000, len = 0
+    ram0   (wx) : org = 0x00000000, len = 0
+    ram1   (wx) : org = 0x00000000, len = 0
+    ram2   (wx) : org = 0x00000000, len = 0
+    ram3   (wx) : org = 0x00000000, len = 0
+    ram4   (wx) : org = 0x00000000, len = 0
+    ram5   (wx) : org = 0x00000000, len = 0
+    ram6   (wx) : org = 0x00000000, len = 0
+    ram7   (wx) : org = 0x00000000, len = 0
+}
+
+/* For each data/text section two region are defined, a virtual region
+   and a load region (_LMA suffix).*/
+
+/* Region to be used for exception vectors.*/
+REGION_ALIAS("VECTORS_FLASH", tcm_vec);
+REGION_ALIAS("VECTORS_FLASH_LMA", tcm_vec);
+
+/* Region to be used for constructors and destructors.*/
+REGION_ALIAS("XTORS_FLASH", ddr_code);
+REGION_ALIAS("XTORS_FLASH_LMA", ddr_code);
+
+/* Region to be used for code text.*/
+REGION_ALIAS("TEXT_FLASH", ddr_code);
+REGION_ALIAS("TEXT_FLASH_LMA", ddr_code);
+
+/* Region to be used for read only data.*/
+REGION_ALIAS("RODATA_FLASH", ddr_code);
+REGION_ALIAS("RODATA_FLASH_LMA", ddr_code);
+
+/* Region to be used for various.*/
+REGION_ALIAS("VARIOUS_FLASH", ddr_code);
+REGION_ALIAS("VARIOUS_FLASH_LMA", ddr_code);
+
+/* Region to be used for RAM(n) initialization data.*/
+REGION_ALIAS("RAM_INIT_FLASH_LMA", ddr_code);
+
+/* RAM region to be used for the banked mode stacks.*/
+REGION_ALIAS("STACKS_RAM", tcm);
+REGION_ALIAS("C1_STACKS_RAM", tcm);
+
+/* RAM region to be used for data segment.*/
+REGION_ALIAS("DATA_RAM", ddr_data);
+REGION_ALIAS("DATA_RAM_LMA", ddr_data);
+
+/* RAM region to be used for BSS segment.*/
+REGION_ALIAS("BSS_RAM", ddr_data);
+
+/* RAM region to be used for the default heap.*/
+REGION_ALIAS("HEAP_RAM", ddr_data);
+
+/* RemoteProc sections at fixed device addresses.*/
+SECTIONS
+{
+    .resource_table : ALIGN(4)
+    {
+        KEEP(*(.resource_table))
+    } > ddr_rsc
+
+    .trace (NOLOAD) : ALIGN(4)
+    {
+        KEEP(*(.trace))
+    } > ddr_trace
+
+    /* Bring-up boot probe, runs from TCM before any DDR execution, the
+       banked mode stacks are allocated after it by rules_stacks.ld.*/
+    .tcm_probe : ALIGN(4)
+    {
+        KEEP(*(.tcm_probe))
+    } > tcm_code
+}
+
+/* Stack rules inclusion.*/
+INCLUDE rules_stacks.ld
+INCLUDE rules_stacks_c1.ld
+
+/* Code rules inclusion.*/
+INCLUDE rules_code.ld
+
+/* Data rules inclusion.*/
+INCLUDE rules_data.ld
+
+/* Memory rules inclusion.*/
+INCLUDE rules_memory.ld
--- a/demos/various/RT-XHAL-T3-GEM-O1/Makefile
+++ b/demos/various/RT-XHAL-T3-GEM-O1/Makefile
@@ -1,0 +1,223 @@
+##############################################################################
+# Build global options
+# NOTE: Can be overridden externally.
+#
+
+# Compiler options here.
+ifeq ($(USE_OPT),)
+  USE_OPT = -O2 -ggdb -fomit-frame-pointer -falign-functions=16
+endif
+
+# C specific options here (added to USE_OPT).
+ifeq ($(USE_COPT),)
+  USE_COPT =
+endif
+
+# C++ specific options here (added to USE_OPT).
+ifeq ($(USE_CPPOPT),)
+  USE_CPPOPT = -fno-rtti
+endif
+
+# Enable this if you want the linker to remove unused code and data.
+ifeq ($(USE_LINK_GC),)
+  USE_LINK_GC = yes
+endif
+
+# Linker extra options here. The ELF entry point is used as the R5F boot
+# vector by the Linux k3-r5 remoteproc driver and must be the vector table.
+ifeq ($(USE_LDOPT),)
+  USE_LDOPT = --entry=_vectors
+endif
+
+# Enable this if you want link time optimizations (LTO).
+ifeq ($(USE_LTO),)
+  USE_LTO = no
+endif
+
+# If enabled, this option allows to compile the application in THUMB mode.
+ifeq ($(USE_THUMB),)
+  USE_THUMB = no
+endif
+
+# Enable this if you want to see the full log while compiling.
+ifeq ($(USE_VERBOSE_COMPILE),)
+  USE_VERBOSE_COMPILE = no
+endif
+
+# If enabled, this option makes the build process faster by not compiling
+# modules not used in the current configuration.
+ifeq ($(USE_SMART_BUILD),)
+  USE_SMART_BUILD = yes
+endif
+
+#
+# Build global options
+##############################################################################
+
+##############################################################################
+# Architecture or project specific options
+#
+
+# Stack size to be allocated to the ARM System/User stack. This stack is
+# the stack used by the main() function before the RTOS is enabled.
+ifeq ($(USE_SYSTEM_STACKSIZE),)
+  USE_SYSTEM_STACKSIZE = 0x800
+endif
+
+# Stack size to the allocated to the ARM IRQ stack.
+ifeq ($(USE_IRQ_STACKSIZE),)
+  USE_IRQ_STACKSIZE = 0x400
+endif
+
+# Stack size to the allocated to the ARM FIQ stack.
+ifeq ($(USE_FIQ_STACKSIZE),)
+  USE_FIQ_STACKSIZE = 0x100
+endif
+
+# Stack size to the allocated to the ARM Supervisor stack.
+ifeq ($(USE_SUPERVISOR_STACKSIZE),)
+  USE_SUPERVISOR_STACKSIZE = 0x100
+endif
+
+# Stack size to the allocated to the ARM Undefined stack.
+ifeq ($(USE_UND_STACKSIZE),)
+  USE_UND_STACKSIZE = 0x100
+endif
+
+# Stack size to the allocated to the ARM Abort stack.
+ifeq ($(USE_ABT_STACKSIZE),)
+  USE_ABT_STACKSIZE = 0x100
+endif
+
+# Enables the use of FPU (no, softfp, hard). The R5F has a VFPv3-D16 FPU.
+ifeq ($(USE_FPU),)
+  USE_FPU = hard
+endif
+
+#
+# Architecture or project specific options
+##############################################################################
+
+##############################################################################
+# Project, target, sources and paths
+#
+
+# Define project name here.
+PROJECT = chibios-xhal-t3-gem-o1
+
+# Target settings.
+MCU = cortex-r5
+
+# Imported source files and paths.
+CHIBIOS  := ../../..
+CONFDIR  := ./cfg
+BUILDDIR := ./build
+DEPDIR   := ./.dep
+
+# Licensing files.
+include $(CHIBIOS)/os/license/license.mk
+# Startup files.
+include $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC/mk/startup_am67r5f.mk
+# OOP and utilities.
+include $(CHIBIOS)/os/common/oop/oop.mk
+include $(CHIBIOS)/os/common/utils/utils.mk
+# XHAL files.
+include $(CHIBIOS)/os/xhal/xhal.mk
+include $(CHIBIOS)/os/xhal/ports/TI/AM67/platform.mk
+include $(CHIBIOS)/os/hal/boards/T3_GEM_O1/board.mk
+# RTOS files (optional).
+include $(CHIBIOS)/os/rt/rt.mk
+include $(CHIBIOS)/os/common/ports/ARMv7-R/compilers/GCC/mk/port_am67.mk
+# Auto-build files in ./source recursively.
+include $(CHIBIOS)/tools/mk/autobuild.mk
+# Other files (optional).
+#include $(CHIBIOS)/os/test/test.mk
+#include $(CHIBIOS)/test/rt/rt_test.mk
+#include $(CHIBIOS)/test/oslib/oslib_test.mk
+
+# Define linker script file here.
+LDSCRIPT = AM67_R5F.ld
+
+# C sources that can be compiled in ARM or THUMB mode depending on the global
+# setting.
+CSRC = $(ALLCSRC) \
+       $(TESTSRC) \
+       main.c
+
+# C++ sources that can be compiled in ARM or THUMB mode depending on the global
+# setting.
+CPPSRC = $(ALLCPPSRC)
+
+# C sources to be compiled in ARM mode regardless of the global setting.
+ACSRC =
+
+# C++ sources to be compiled in ARM mode regardless of the global setting.
+ACPPSRC =
+
+# C sources to be compiled in THUMB mode regardless of the global setting.
+TCSRC =
+
+# C++ sources to be compiled in THUMB mode regardless of the global setting.
+TCPPSRC =
+
+# List ASM source files here.
+ASMSRC = $(ALLASMSRC)
+
+# List ASM with preprocessor source files here.
+ASMXSRC = $(ALLXASMSRC)
+
+# Inclusion directories.
+INCDIR = $(CONFDIR) $(ALLINC) $(TESTINC)
+
+# ARM-specific options here.
+AOPT = -marm
+
+# THUMB-specific options here.
+TOPT = -mthumb -DTHUMB
+
+# Define C warning options here.
+CWARN = -Wall -Wextra -Wundef -Wstrict-prototypes
+
+# Define C++ warning options here.
+CPPWARN = -Wall -Wextra -Wundef
+
+#
+# Project, target, sources and paths
+##############################################################################
+
+##############################################################################
+# Start of user section
+#
+
+# List all user C define here, like -D_DEBUG=1
+# The core feature set comes from the AM67R5F startup device, it is not
+# overridable from here.
+UDEFS =
+
+# Define ASM defines here
+UADEFS =
+
+# List all user directories here
+UINCDIR =
+
+# List the user directory to look for the libraries here
+ULIBDIR =
+
+# List all user libraries here
+ULIBS =
+
+#
+# End of user section
+##############################################################################
+
+##############################################################################
+# Common rules
+#
+
+RULESPATH = $(CHIBIOS)/os/common/startup/ARMCRx/compilers/GCC
+include $(RULESPATH)/mk/arm-none-eabi.mk
+include $(RULESPATH)/mk/rules.mk
+
+#
+# Common rules
+##############################################################################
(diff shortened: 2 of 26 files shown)


gdisirio:
Thanks, this is a substantial and well-documented foundation. The board, VIM and system-timer portions look coherent, but I found several issues in the UARTv1 SIO LLD that should be addressed before merge. The polling-only loopback test would not expose these interrupt-path problems.

- sio_lld_start() tests for HAL_DRV_STATE_STOP, but drvStart() changes the state to HAL_DRV_STATE_STARTING before invoking the LLD. The condition is therefore never true and the VIM handler is never registered or enabled. Similarly, drvStop() changes the state to HAL_DRV_STATE_STOPPING before sio_lld_stop(), so its READY test prevents all shutdown work. The straightforward fix is to perform activation and teardown directly in these transition methods, or test STARTING and STOPPING respectively.
- sio_lld_start() discards the pointer returned by sio_lld_setcfg(). With a NULL configuration, the default is programmed but siop->config stays NULL, violating the base-driver postcondition. Please assign the returned pointer to siop->config and undo activation if configuration is rejected, following the RP UART LLD pattern.
- Reject config->baud == 0U before calculating the divisor; currently this divides by zero. It would also be prudent to mask TI_UART_LCR_DLAB out of the stored line configuration because leaving DLAB set redirects THR and IER accesses to the divisor latches.
- The driver says every LSR read must pass through uart_latch_lsr(), but sio_lld_is_rx_empty(), sio_lld_is_rx_idle(), sio_lld_is_tx_ongoing() and the THRE handler read LSR directly. Those paths can consume an overrun indication without latching it. Please route those reads through the helper, likely by making the state checks small LLD functions instead of direct-register macros.
- sio_lld_get_and_clear_events(siop, events) returns only requested events but clears all SIO_LSR_STICKY bits. It should clear only the sticky conditions represented by events, leaving unrequested errors and RX-idle pending.
- The TI UART FCR receive-trigger encodings are 8, 16, 56 and 60 characters. The current names 1, 4, 8 and 14 are standard-16550 values; in particular RXTRIGGER_8 currently programs a 56-character threshold. Renaming the constants and making RXTRIGGER_8 use encoding zero is sufficient.
- ETBEI reports THR/FIFO empty, not transmission complete. If the THRE interrupt arrives while TEMT is still clear, the handler masks ETBEI and never arranges another notification, so sioSynchronizeTXEnd() can remain suspended. This needs a follow-up mechanism, such as the polling virtual timer used by the RP UART driver, rather than assuming THRE and TEMT become true together.

Could you update these points and add an interrupt-driven RX/TX synchronization test if the board setup permits it?

arasnazmi:
Thanks for the review. All seven points are addressed, one commit each

gdisirio (changes requested):
I am Codex, Giovanni Di Sirio's AI assistant, posting this follow-up review at his request.

I reviewed the complete PR at e76dd893abcbc56ce45152d84fa33edd555ee85a, including the UART fixes and new self-test. I also checked the subsequent commit 7505e6d5b87aa5276f4d966b1b383247cee7bcea: it only adds copyright notices, so the findings below still apply.

The previous corrections are substantially present, but I recommend addressing these remaining issues before merging:

1. **[P1] Missing IRQ entry/exit bookkeeping.** In [port_platform.c](https://github.com/chibios-upstream/chibios/blob/7505e6d5b87aa5276f4d966b1b383247cee7bcea/os/common/ports/ARMv7-R/platforms/am67/port_platform.c#L103), the dispatcher invokes handlers without CH_IRQ_PROLOGUE()/CH_IRQ_EPILOGUE(). Neither the handlers nor the assembly entry supplies that bookkeeping. With CH_DBG_SYSTEM_STATE_CHECK=TRUE, the first tick reaches chSysLockFromISR() with isr_cnt == 0 and halts with SV#6. IRQ hooks, statistics and tracing are also skipped. Please bracket dispatch with the appropriate kernel IRQ entry/exit macros, keeping the existing preemption-return convention.

2. **[P1] RX character timeout can block an unrelated TX waiter.** The [CTI handler](https://github.com/chibios-upstream/chibios/blob/7505e6d5b87aa5276f4d966b1b383247cee7bcea/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L810) disables the entire UART vector. Consider a short incoming burst timing out while a longer transmission is underway, before its THRE interrupt has armed the TX-end timer. The later TX interrupt cannot run. A thread waiting for TX completion before draining RX can time out, or remain blocked with TIME_INFINITE. Taking the reported hardware CTI behavior as given, the workaround still needs to preserve TX progress while RX is unread; simply masking the shared vector is insufficient.

3. **[P1] Event acknowledgement can repeatedly re-arm idle TX interrupts.** [sio_lld_get_and_clear_events()](https://github.com/chibios-upstream/chibios/blob/7505e6d5b87aa5276f4d966b1b383247cee7bcea/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L613) calls the global enable-update function, which reconstructs all IER enables and unconditionally enables the VIM line. An acknowledgement-only callback with TX events enabled therefore undoes the THRE handler's masking even though no new TX data was written. With the FIFO still empty, the interrupt repeats. The same global update can undo the CTI controller mask before RX has been drained. Please use source-specific re-arming and preserve unrelated service masks; acknowledging RX errors/events should not re-arm idle TX.

4. **[P2] Live configuration leaves interrupts disabled.** [sio_lld_setcfg()](https://github.com/chibios-upstream/chibios/blob/7505e6d5b87aa5276f4d966b1b383247cee7bcea/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L481) clears IER and does not restore it. Initial start subsequently applies the enabled-event mask, but drvSetCfgX()/drvSelectCfgX() do not. A successful live configuration therefore leaves siop->enabled populated while the actual interrupt enables are zero. A subsequent RX wait can time out despite incoming data. Please restore the appropriate interrupt state when configuring an already-started driver.

5. **[P2] A historical RX-idle event is used as current activity state.** [sio_lld_is_rx_idle()](https://github.com/chibios-upstream/chibios/blob/7505e6d5b87aa5276f4d966b1b383247cee7bcea/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L332) treats the sticky CTI bit as proof that RX is currently idle. Draining the FIFO does not clear that bit, so during a subsequent burst the query still returns true until the application explicitly acknowledges the old event. This also makes sioSynchronizeRXIdle() return immediately for the wrong receive cycle. Please separate an unconsumed historical event from current activity; simply discarding the historical event on read would undermine event retention.

Validation: the demo builds cleanly with GCC 14.3.1 in both its default configuration and with checks, assertions and the system-state checker enabled. A host harness compiled the actual UART LLD against mocked registers and kernel/controller operations. It demonstrated the disabled IER after live configuration, repeated TX re-arming through an acknowledgement-only callback, the CTI path masking TX without arming its timer, and stale RX-idle state across bursts. These are software-model results, not hardware reproductions; the interrupt-storm model explicitly assumes THRE assertion when enabled with an empty FIFO. The debug-entry finding follows from the kernel checks and the compiled interrupt path; I have not run the debug build on a board.

Suggested regression tests: debug-enabled startup; overlapping short RX and longer TX with TX-end awaited before RX is drained; an acknowledgement-only callback; live configuration followed by interrupt-driven RX; and two receive bursts without consuming the first burst's idle event. The new loopback test is useful, but does not cover all these cases.

I found no additional merge blocker in the board/linker setup or periodic timer configuration during this pass.

gdisirio:
Follow-up from Codex, Giovanni Di Sirio’s AI assistant, at his request: please remove the added T3 Foundation copyright line from `demos/various/RT-XHAL-T3-GEM-O1/cfg/xhalconf.h`. This is a generated configuration file; its header is controlled by the generator template, and a manual copyright addition would be overwritten on regeneration. Please leave the generated header unchanged. This request is specific to `xhalconf.h`, not a rejection of the copyright additions in authored source files.

arasnazmi:
Thanks, all are fixed

gdisirio:
Follow-up from Codex, Giovanni Di Sirio's AI assistant, at his request.

Thanks for the updates. I checked the changes through 1ae5c140c3ffcaec9230b95212130c1ce8880964. The IRQ bookkeeping correction and generated-header copyright cleanup look good. However, three UART cases still need attention:

1. **RX-idle synchronization with unread data.** sio_lld_is_rx_idle() now checks only whether the FIFO is empty. If a thread waits in sioSynchronizeRXIdle() while data is buffered, CTI wakes it, but the synchronization loop checks the FIFO again, finds the unread data, and suspends again even though reception has ended. The earlier request was to distinguish current receive-cycle state from an unconsumed historical event, not to equate line idle with FIFO empty.

2. **TX starting after CTI has masked the vector.** The CTI handler starts the polling timer only when TX is already busy. If TX is idle at that moment and the application subsequently starts a transmission without first draining RX, the vector remains disabled, uart_arm_tx() only writes IER, and no polling timer is started. A TX-space or TX-end waiter can therefore still time out or hang. Progress needs to be maintained for transmissions started while the vector is masked, not just those already in flight.

3. **Live reconfiguration after CTI masking.** sio_lld_setcfg() now restores IER, but leaves the VIM enable untouched. If CTI previously disabled the vector, configuration resets the RX FIFO and removes the reason for the mask, yet the vector remains disabled. Subsequent interrupt-driven reception can stall until some separate operation re-enables it.

These follow-up findings are from code inspection, not new hardware reproductions. The green CI checks do not exercise these receive/transmit sequences.

Would you like help working through the remaining state handling and preparing a patch? Giovanni and I can help with implementation, cross-builds and software/register-model checks, but we do not have this board available and cannot test on hardware. We would need you to validate the resulting patch on the T3 Gemstone O1, especially these three sequences.

arasnazmi:
All three are right, and all three came from the same mistake: I was inferring the mask and the receive-cycle state instead of recording them. Both are driver state now.

| Finding | Commit |
|---|---|
| 1. RX-idle synchronization with unread data | `1c04b878` |
| 2. TX starting after CTI has masked the vector | `577e31de` |
| 3. Live reconfiguration after CTI masking | `577e31de` |

`rx_idle` is the state of the current receive cycle: set when the line goes quiet, cleared as soon as a frame arrives. `sio_lld_is_rx_idle()` reports an empty FIFO or a quiet line, so `sioSynchronizeRXIdle()` no longer falls back into its loop with data buffered, while the latched event stays separate and does not carry into the next burst.

`rx_masked` records that the vector is down. `uart_arm_tx()` hands a transmission started while it is up to the polling timer, so a TX-space or TX-end waiter is served whether or not the transmission was already in flight, and `sio_lld_setcfg()` drops the mask and brings the line back, since resetting the FIFOs destroys the unread frames the mask existed for.

Verified on the T3 Gemstone O1, with `main_uart1` reserved from Linux so the firmware is the only owner of the port and of the interrupt:
```
uart: reserved
irq:  not registered on the host side
UART1: on=1 clock=1 state=2/1 fck=48000000 resets=00000000
SIO selftest passed
```

Both configurations build clean, default and with checks, assertions and the system-state checker enabled.

Thanks for the offer the useful split here is the other way round. The board is the part we have, so validation is not the bottleneck; and since each of these findings took one edit and one board cycle, routing them through a patch
exchange would be slower than the review loop we already have. Please keep inspecting: three rounds have now found things the hardware test could not, including two I introduced myself.

Still open from the previous round: with the state checker enabled the demo halts at `SV#4`, `chSysLock()` with the lock already held. It reproduces with the SIO driver removed entirely — kernel, tick and one sleeping thread — so it is not the UART, but I cannot yet say whether it is the AM67 sub-port or the common ARMv7-R code, since the minimal case still runs through our VIM dispatcher and DMTIMER tick handler. I am taking it, and I will narrow that down before proposing anything. 
One question while I do, because it decides where I look first. In `chcoreasm.S` the preemption return brackets `chSchDoPreemption()` with `__dbg_check_lock()` and `__dbg_check_unlock()`. That pairing holds if the incoming thread always resumes at a point that performs the matching unlock. Is that invariant meant to cover a thread entering through `__port_thread_start()` as well, or should the trampoline's own `__dbg_check_unlock()` be the only one in that path? I would rather understand the intent than pattern-match a fix onto it.

gdisirio:
Follow-up from Codex, Giovanni Di Sirio's AI assistant, at his request. I read your latest reply and reviewed through `577e31de`. Thanks for the board validation; continuing with review here and hardware validation on your side makes sense.

The live-reconfiguration fix now restores the vector correctly, and TX started after CTI now progresses from ordinary unlocked thread context. Two UART corner cases remain. These are separate from your `SV#4` reproduction with SIO removed.

### 1. Thread-only timer API reached from X-class writes

[`uart_arm_tx()` calls `chVTSet()`](https://github.com/chibios-upstream/chibios/blob/577e31de6ea7384926d8110c3dcdbf0737465d96/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L277), but `sioAsyncWriteX()` and `sioPutX()` may be called from ISR/callback or already-locked context. For example, a callback responding to CTI can write while `rx_masked` is true. That reaches the thread-only `chSysLock()` inside `chVTSet()`, which the state checker rejects with `SV#4`. A write from an already-locked context has the same problem.

A relatively small owner-safe approach is to start the existing polling timer whenever CTI masks the vector, even if TX is currently idle, and keep it alive while the vector remains masked. The write paths can then mark/enable TX work without manipulating the timer. Keep TX wakeups/callbacks gated by pending work or completion transitions, so an idle transmitter does not generate a callback every timer tick; preserve any TX-end polling still needed when the RX mask is released.

If you retain write-side timer arming instead, it needs an X-context-safe locking solution, not simply `chVTSetI()` without a lock. Note that this ARMv7-R port does not currently expose `port_get_lock_status()` / `port_is_locked()`, so the `chSysGetStatusAndLockX()` pattern used elsewhere is not a drop-in solution here. The existing header comment saying the timer is never armed from write paths also needs to match the chosen implementation.

### 2. RX-idle cycle state survives a drained FIFO

[`rx_idle` is cleared by RDA](https://github.com/chibios-upstream/chibios/blob/577e31de6ea7384926d8110c3dcdbf0737465d96/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L950), not by the arrival of every frame. With the configured eight-frame threshold:

- CTI sets `rx_idle`.
- The application drains RX and the vector is re-enabled, but `rx_idle` remains true.
- A new burst has fewer than eight frames, so no RDA has occurred yet. `sioIsRXIdleX()` still returns true, and `sioSynchronizeRXIdle()` can return before this burst's timeout.

This happens even if the application acknowledged the previous RX-idle event. A small fix for this sequence is to clear `rx_idle` in the empty-FIFO re-arm paths, before unmasking the vector (`uart_arm_rx()` and the equivalent branch in `sio_lld_update_enable_flags()`). Leave `SIO_LSR_CTI` alone there, preserving the historical event. Then an empty FIFO still reports idle, while a newly nonempty FIFO must wait for its own CTI instead of inheriting the previous cycle's flag. Please include a second short burst, sampled before CTI, in the regression test.

### ARMv7-R thread-start / preemption question

Yes, the lock-handoff invariant includes a newly started thread. Its trampoline performs the unlock for that path; the unlock after `chSchDoPreemption()` is not also executed on entry to that new thread.

`__port_switch_arm()` switches stacks and restores the incoming context's saved return address. For a new thread, `port_setup_context()` sets that address to `__port_thread_start()`. The outgoing thread's continuation after `chSchDoPreemption()` remains on its own saved stack and executes only when that thread is scheduled again, under the lock. A previously blocked thread instead resumes its own saved kernel continuation and reaches that path's unlock. I would therefore keep both assembly checks rather than remove one to compensate for the failure.

Before changing the port, please verify that `CH_DBG_SYSTEM_STATE_CHECK` has the same value in C and assembly, with a clean rebuild. This makefile has separate `UDEFS` and `UADEFS`; a C-only override (`USE_COPT` or `UDEFS` alone) can enable the C checks while leaving the conditional assembly lock/unlock bookkeeping absent. That can produce a stale lock count even with SIO removed. This is a diagnostic possibility, not a diagnosis of your hardware run.

I rebuilt with the debug defines supplied to both `UDEFS` and `UADEFS` and verified in the disassembly that the preemption lock/unlock pair and the trampoline unlock are present. If your image already has those instructions, please capture the failing `__dbg_check_lock()` call site/return address, CPSR, `isr_cnt`, and `lock_cnt`. After the outermost `CH_IRQ_EPILOGUE()` and before the assembly preemption lock check, both counters should be zero; an incoming thread reaches its unlock with `isr_cnt == 0` and `lock_cnt == 1`. That should distinguish an unbalanced IRQ handler from a context-switch/configuration problem without guessing from the shared `SV#4` label.

Validation here: the actual UART LLD compiled against mocked registers/kernel operations confirms the two UART paths above, as well as the corrected normal-thread TX and live-reconfiguration paths. Default and consistently debug-enabled ARM builds pass. These remain software-model/build results, not hardware reproductions.

arasnazmi:
You were right about the build, and that retracts my ARMv7-R report entirely. I had passed the debug defines through `UDEFS` only. Disassembling the images I had actually run: zero `__dbg_check_lock` / `__dbg_check_unlock` calls in
`Irq_Handler` and `__port_thread_start`. The C checks were live while the assembly never did its half of the bookkeeping — exactly the stale lock count you described. Rebuilt with the defines in both `UDEFS` and `UADEFS`, two calls in `Irq_Handler` and one in the trampoline, and the debug build simply runs. There is no `SV#4` and no port bug. Please disregard that report; it was my measurement, not your code, and thank you for not taking it at face value.

The two UART cases are fixed. Seven commits in total, the other five are explained below.

| Commit | |
|---|---|
| `117a07c2` | 1. Thread-only timer API reached from X-class writes |
| `5cf59569` | 2. RX-idle cycle state survives a drained FIFO |
| `8a1fdb1f` | The driver did not build without virtual timers |
| `536df4f6` | A character timeout survives configuration |
| `95831e71` | A break-only enable request leaves the interrupt off |
| `226904f1` | Regression sequences for the three review findings |
| `40816283` | Coverage for errors, restart and transmitter space |

I took your first suggestion. The handler starts the timer whenever it masks the vector, not only for a transmission already in flight, and keeps it armed while the mask is up; the write paths no longer touch it, which makes the
header's claim about where it is armed true again. Since it now outlives the transmission, its reports are gated on outstanding work. The receive cycle flag is cleared in the empty-FIFO re-arm paths, `SIO_LSR_CTI` untouched.

Then I wrote the regression sequences you asked for, and went back over the three paths I had never entered at all: line errors, stop and restart, and waiting for transmitter space. Between them they found three more driver bugs.

`8a1fdb1f` — the driver carried a second field set for a build without virtual timers and did not compile: `txend_done` was only in the RT set. Making it build would not have made it work, since the timer is what carries the transmitter while the vector is masked, so without it a timeout stops transmission until the receiver is drained. That is worse than the RP driver's degradation, so the dependency is now stated outright and the dead field set is gone.

`536df4f6` — the live reconfiguration case still failed, with reception dead afterwards. Resetting the FIFOs does not retire the character timeout: it stands until the receive register is read and survives a read of IIR, and nothing was reading either because the vector was masked for it. Enabling the line at the end of configuration delivered the old timeout, the handler masked again, this time against an empty receiver where no read can arrive to lift it. Configuration now retires what the receiver has standing, with the enables still down, and only when the peripheral says something is standing — reading an empty receive register speculatively broke an unrelated transfer, which the
test caught too. 

`95831e71` — the last one is from the coverage I had left out rather than
from the sequences above. A break is reported as a receiver error, the way
the RP driver does it: the status bit is latched, returned and consumed with
the error bits, and a pending break brings `sioSynchronizeRX()` back with
`SIO_MSG_ERRORS`. But the enabled set was tested against `SIO_EV_ALL_ERRORS`,
which does not contain `SIO_EV_RX_BREAK`, so an application asking for breaks
and nothing else left `ELSI` disabled: the break was then only findable by a
poll, never by a thread waiting on the port.

Verified on the T3 Gemstone O1 with `main_uart1` reserved from Linux — node
`status = "reserved"`, no tty bound, no host interrupt registered — in both
configurations, the debug one with the defines reaching the assembly:

```
RT-XHAL-T3-GEM-O1 starting
  pass6: break errors=1000
  pass6: overrun errors=08
SIO selftest passed
```

The self-test now covers, on top of the two original transfers: a
transmission started under the mask; repeated acknowledgement with an idle
transmitter; configuration applied while unread, followed by interrupt-driven
reception; two receive cycles with the first idle event unconsumed and the
second sampled before its own timeout; a break and an overrun, both produced
on the port itself, with the recovery that has to follow them; the driver
stopped and started again from the state with the most standing in it; and a
wait for transmitter space, with the vector live and with it masked. Each
failure names its step.

The break-only case was confirmed the honest way round: with the fix reverted
the board reports `SIO selftest FAILED at break-enable`, `ier=00`.

That is the part I should have had earlier. Every round since the first has
found something the two-transfer test could not see, including this one, and
the three areas I had left untested turned out to hold one more.

gdisirio:
Follow-up from Codex, Giovanni Di Sirio's AI assistant, at his request. I completed the final review at `40816283a3163fe1a02ba840965f06ec24252f5b`.

The RT-only requirement is acceptable to Giovanni for this device. The previous two UART fixes pass the regression checks below, and the withdrawn ARMv7-R report is no longer an open point. Three findings remain:

### 1. [P1] Startup enables the vector before installing its handler

[`sio_lld_start()` calls `sio_lld_setcfg()` before registering the handler](https://github.com/chibios-upstream/chibios/blob/40816283a3163fe1a02ba840965f06ec24252f5b/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L523), but [configuration now unconditionally enables the VIM vector](https://github.com/chibios-upstream/chibios/blob/40816283a3163fe1a02ba840965f06ec24252f5b/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L673). The same ordering occurs on restart, after stop has removed the handler.

If the UART raises a level-sensitive source during that interval, the dispatcher has no handler to acknowledge the peripheral; acknowledging only the controller allows the still-asserted source to re-enter before the starting thread can finish. Taking the documented CTI behavior as given, this remains relevant with IER cleared: data can arrive after UART mode is enabled, including while the starting thread is preempted.

Please keep the vector masked through initial configuration and handler/priority installation. Configuration should restore the vector only for an already registered/live driver, and the polling interval should be initialized before unmasking. The register model confirms the enable-without-handler ordering; the IRQ-storm consequence is from code inspection, not a hardware reproduction.

### 2. [P2] Polling notifications ignore the enabled-event mask

[`uart_txend_timer_cb()`](https://github.com/chibios-upstream/chibios/blob/40816283a3163fe1a02ba840965f06ec24252f5b/os/xhal/ports/TI/LLD/UARTv1/hal_sio_lld.c#L324) wakes TX/TX-end waiters and invokes the registered callback without consulting `siop->enabled`.

Sequence: CTI starts the persistent timer; disable all events; write while RX remains unread; let TX complete. IER stays zero, but the polling path still delivers notifications. The model records `enabled=0 ier=0 callbacks=1 tx=2 txend=1` with a registered acknowledgement-only callback.

Please preserve the polling needed while the vector is masked, but gate notifications by the corresponding enabled events. This should not reintroduce thread-only timer calls into X-class write paths.

### 3. [P2] The echo loop cannot recover from receive errors

The [demo echo loop](https://github.com/chibios-upstream/chibios/blob/40816283a3163fe1a02ba840965f06ec24252f5b/demos/various/RT-XHAL-T3-GEM-O1/main.c#L609) handles only `MSG_OK`. After an overrun, framing/parity error or break, `sioSynchronizeRX()` returns `SIO_MSG_ERRORS` without consuming the latched error. The loop never acknowledges it or drains the affected data, so subsequent waits return immediately forever: echo stops and the main thread busy-spins.

A bounded test compiling the actual echo loop and generic RX synchronizer recorded 32 iterations, zero blocking waits, and the error still latched. Please handle `SIO_MSG_ERRORS` explicitly, acknowledging errors and draining/discarding affected data as the self-test recovery does.

### Validation

- Normal and consistently debug-enabled ARM demo builds pass without warnings with GCC 14.3.1. Debug definitions reach both C and assembly; disassembly confirms the preemption lock/unlock pair and trampoline unlock.
- All 19 added C/header files pass the style checker; `git diff --check` passes.
- Eight regression groups pass with the actual UART LLD compiled against mocked kernel/controller operations, using AddressSanitizer and UndefinedBehaviorSanitizer. These include callback/locked-context writes, TX polling across RX unmask, idle notification suppression, selective error acknowledgement, invalid-configuration rollback and restart state.
- No hardware testing was performed here. The model supplies hardware progress at explicit boundaries; it does not emulate TI register read side effects, physical CTI behavior or board timing.

arasnazmi:
Thanks for feedback

### 1. [P1] Startup enables the vector before installing its handler — fixed in `da3f938d0b`

`sio_lld_setcfg()` now only restores the vector for an already-live driver (`siop->state == HAL_DRV_STATE_READY`); `sio_lld_start()` unmasks it itself, after `vimSetHandler()`/`vimSetPriority()` have run. The polling interval is computed before the vecte serviced synchronouslyon unmask always has a valid `txend_step` if the handler starts the timer. 
### 2. [P2] Polling notifications ignore the enabled-event mask — fixed in `5267fff9d2` `uart_txend_timer_cb()` akeup and the TX-endwakeup/callback on `uart_tx_ier(siop) != 0`. The TX-end transition (`TEMT` latch) is still tracked g events later doesn'trediscover a stale transition.

### 3. [P2] The echo loop cannot recover from receive errors — fixed in `c52e61e84a`

The echo loop now handle:`sioGetAndClearErrorsX()` acknowledges the latched error and the receiver is drained, mirroring the rlready used.

### Hardware validation

Built and flashed to the(J722S MCU R5F,`79000000.r5f`), GCC 14.3.1, no warnings. `remoteproc` trace (`/sys/kernel/debug/remolean boot:

RT-XHAL-T3-GEM-O1 starti
  pass6: break errors=1000
  pass6: overrun errors=
SIO selftest passed
heartbeat 0 t=49 ms
...

All self-test passes (1–8, including pass6 break/overrun, pass7 restart, pass8 tx-space) pass; heartbearward with no crash, nowatchdog reset, no anomalies in `dmesg`. Findings #1 and #2 are covered by that run (their previousas a startup IRQ storm or spurious notifications during pass6/pass8); finding #3's new echo-loop branch  was verified by build + ive line-errorinjection, since exercising it needs external fault injection on the UART line rather than just a clean

gdisirio (approved):

Follow-up from Codex, Giovanni Di Sirio's AI assistant, at his request.

Scope: TI AM67/J722S Cortex-R5F XHAL platform, T3 Gemstone O1 board and demo, targeting `master`. Re-reviewed at `c52e61e84a7c84dbda9e9f316d4e3f5dd3e7d951`.

Stage reached: D (style), following verification of the remaining correctness findings.

All three findings from the previous review are resolved:

1. Startup/restart installs the interrupt handler and priority, and initializes the polling interval, before unmasking the vector. Live configuration still restores the vector for a READY driver.
2. TX polling notifications now honor the transmitter-event enable state. Polling still carries TX while unread RX holds the shared vector masked; disabling TX events suppresses notifications, and re-enabling them does not replay an already observed completion.
3. The echo loop acknowledges `SIO_MSG_ERRORS` and drains the affected receive data, removing the sticky-error busy loop.

The earlier fixes remain intact, including ISR/locked-context writes without thread-only timer calls. The withdrawn ARMv7-R debug report is closed; consistently enabling the C and assembly checks requires no port change.

Mechanical gates and validation:

- Normal and debug ARM demo builds pass without warnings. Debug definitions reach both C and assembly; disassembly confirms the IRQ lock/unlock pair and the thread-trampoline unlock.
- Targeted tests compiling the actual UART LLD against mocked registers, kernel operations and interrupt-controller operations pass with AddressSanitizer and UndefinedBehaviorSanitizer. They cover startup/restart ordering, live reconfiguration, five event-mask combinations, ISR/locked writes, TX-space/TX-end progress, idle notification suppression, re-enabling events, RX unmask and timer teardown. Leak detection was disabled because it is unsupported in this execution sandbox.
- Changed C/header files pass the style checker; `git diff --check` passes. All current CI checks are green.
- No hardware testing was performed here. The register model does not emulate TI register read side effects or physical UART timing; board validation remains the author's reported testing. The echo recovery change was checked by inspection and compilation in this pass.

Verdict: advisory-LGTM. This approval supersedes my earlier request for changes; the merge decision remains Giovanni's.

Changelog housekeeping, not a review finding: I prepared the following entry for `CHANGELOG.md` under Unreleased / Added, but could not push it because the fork has maintainer edits disabled and our account has no direct write access. It remains a maintainer-side follow-up:

```markdown
- [XHAL] TI AM67/J722S Cortex-R5F platform and T3 Gemstone O1 board support,
  including VIM interrupt dispatch, the DMTIMER system timer, a UART SIO
  driver, and a Linux remoteproc demo with UART loopback self-tests
  ([#289](https://github.com/chibios-upstream/chibios/pull/289)).
```

PR #290, opened 2026-09-04
SDC: MMC/eMMC cards expect RCA to be assigned by host

dron0gus:
## Summary

eMMC (and modern MMC card, but not tested) expect host to set Relative Card Address using CMD3. While SD card randomly generates RCA address and provides it to host with CMD3 reply.

## Related

- Issue(s):
- Notes for reviewers:

## Target Branch Check

- [ ] This PR targets `master` — all changes land on `master` first (upstream-first policy);
      merged on `master`

## Scope

- [ ] I reviewed impact on other ports/configurations where relevant

## Mechanical Checks

- [ ] Style check passed on changed `.c`/`.h` files

## Testing

- [x] Tested on custom STM32F767 device with KLMAG1JETD-B041 eMMC chip.

## Compatibility and Risk

### DIFF
--- a/os/hal/src/hal_sdc.c
+++ b/os/hal/src/hal_sdc.c
@@ -711,10 +711,20 @@ bool sdcConnect(SDCDriver *sdcp) {
     goto failed;
   }
 
-  /* Asks for the RCA.*/
-  if (sdc_lld_send_cmd_short_crc(sdcp, MMCSD_CMD_SEND_RELATIVE_ADDR,
-                                 0, &sdcp->rca)) {
-    goto failed;
+  if ((sdcp->cardmode &  SDC_MODE_CARDTYPE_MASK) == SDC_MODE_CARDTYPE_MMC) {
+    /* Set RCA.*/
+    /* MMC card expects its address in high 16 bits of argument */
+    sdcp->rca = 0x0002 << 16;
+    if (sdc_lld_send_cmd_short_crc(sdcp, MMCSD_CMD_SET_RELATIVE_ADDR,
+                                   sdcp->rca, resp)) {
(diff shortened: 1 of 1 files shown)


gdisirio:
Thanks, this change makes sense. It looks like the same issue is present in XHAL: sdcConnect() in os/xhal/codegen/hal_sdc.xml still unconditionally sends CMD3 with argument zero and stores the response as the RCA, and os/xhal/src/hal_sdc.c mirrors that generated implementation. Do you agree that XHAL is affected too? If so, could you update the XML and regenerate the source as part of this PR?

Also, MMC CMD3 returns an R1 response. Would it make sense for the new MMC branch to check MMCSD_R1_ERROR(resp[0]) in addition to the low-level command result, so card-reported errors are not accepted silently?

dron0gus:
Sorry I did not look at maser. We are still on stable-21.11. So I simply ported this fix to master and PR. And I did not check os/xhal/codegen/hal_sdc.xml in master.
I'll take a closer look.

Yes, it makes sense to check response. I'm just bringing up new HW and this is straight fix I have now. I need to check mmc spec in details...

gdisirio:
I can port to XHAL and 21.11.x afterward, just look into the other issue, I cannot test with MMC.

gdisirio:
Following up on behalf of Giovanni as his AI assistant: we will merge this PR as-is. The additional CMD3 R1 status check we suggested is not a merge requirement; we have not identified a concrete failure demonstrating that its absence causes a problem. We can revisit it if such evidence emerges. Giovanni will handle the XHAL port and the 21.11.x backport afterward. Thanks for the fix!

PR #320, opened 2026-09-16
TI AM67: add the MCSPIv1 SPI driver

arasnazmi:
## Summary

Adds the AM67 MCSPI driver to the TI XHAL port (`os/xhal/ports/TI/LLD/MCSPIv1/`),
following the layout the ST and SIO drivers already established, with instance
parameters coming from `am67_registry.h` rather than being hardcoded.

The XHAL `setcfg()`/`selcfg()` channel model removes a defect the classic driver shape
had. Re-running the whole controller bring-up, soft reset included, on every chip
select change was measured at roughly 150 controller resets per second with two
devices sharing the bus, which reset an LPS22DF out from under its own driver. The
XHAL split lets the one-time reset stay in `spi_lld_start()` while a channel switch
only reprograms `CHCONF`/`CHCTRL`.

A configuration is validated before it is programmed, following
`spi_lld_validate_config()` in the RP and STM32 ports: circular and slave modes,
frame sizes other than 8 bits, a chip select past channel 3, and a clock the
`CLKD`/`EXTCLK` pair cannot divide down to are all refused rather than masked or
clamped into range. `CHCONF.WL` is hardwired to 8 bits here, so accepting a wider
frame size would make `spiGetFrameSizeX()` disagree with what the hardware shifts.

`am67_padcfg.h` arrives with this PR as the AM67 pad-configuration helper. It sits in
the family directory rather than in the MCSPI LLD, because pad configuration is
SoC-level and the drivers that follow need it too.

Board knowledge stays out of the driver where it can: which device sits behind which
chip select, and the GPIO line that releases the onboard IMU, live in the board files
as `board_imu_enable()`, which is why this PR also touches `board.c`/`board.h`.

The one exception is the SPI0 pinmux, which the driver applies in `spi_lld_start()`
in case the Linux device tree leaves the pads unconfigured; the write is idempotent
otherwise. That set includes `MCU_MCAN0_TX`, which carries `SPI0_CS3` as an alternate
function on this SoC, so a board using MCU_MCAN0 would lose its CAN TX pin when the
SPI driver starts. `SPI0_CS2` is deliberately left to the board for the same class of
reason — it borrows the wakeup-domain UART0 pad. Both choices are commented in the
driver and are the part of this PR most worth a reviewer's eye.

## Related

- Issue(s): none
- Notes for reviewers: builds on the AM67 XHAL foundation PR (#289), already merged,
  and is based directly on `master`. The demo gains a WHO_AM_I probe so the driver is
  exercised on hardware.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

      all 8 changed `.c`/`.h` files: no findings
      `RT-Posix-Simulator` links, the ARM targets in `test/rt-ports/testbuild` build
      (ARMv6-M-ALT, ARMv7-M, ARMv7-M-ALT, ARMv8-M-ML, ARMv8-M-ML-ALT), and
      `RT-XHAL-T3-GEM-O1` builds warning-free at text 20280 B with
      `arm-none-eabi-gcc 13.2.1`

## Testing

- Any other tests run on a device? Yes. On a T3 Gemstone O1 (AM67A/J722S) R5F the
  demo's probe read `WHO_AM_I` from an ICM-20948 over `MCU_MCSPI0` channel 3 and got
  back `0xEA`, the device's documented value, on 20 of 20 power-on boots. That
  confirms the select / polled exchange / unselect path end to end against a real
  device rather than a register model.

## Compatibility and Risk

A new driver under a new path; no existing driver is modified.

The hardware test covers the polled path only. The interrupt-driven transfer path —
`spi_lld_exchange()`/`send()`/`receive()`/`ignore()` and the ISR — is compile-verified
but has not been exercised on a device, because the demo's probe is a single register
read and the XHAL SPI class exposes no polled entry point of its own. A
transfer-driven demo is the obvious next step and I am happy to add one before this
merges. The driver has no DMA support. Multi-master arbitration is likewise untested,
as is operation with more than one device on the bus.

`board_imu_enable()` drives MCU_GPIO0 directly rather than through a PAL driver: this
port has no PAL yet. It is a board-file helper the application calls explicitly, not
something the SPI driver does behind the caller's back.


gdisirio (changes requested):
Thanks, the overall direction and the separation between controller startup and live channel configuration make sense. I see several issues that should be addressed before merging, however:

1. `spi_lld_unselect()` clears `CHCONF.FORCE` immediately. RX full does not guarantee that the last frame has reached wire EOT, so an immediate unselect can truncate the final word. The cited NuttX implementation explicitly waits for `CHSTAT.EOT` before deasserting CS for this reason. The 1 MHz WHO_AM_I probe may not expose this race.

2. `spi_lld_stop_transfer()` masks the interrupt, clears status, and drains RX, but it does not stop or complete a word already shifting. That word can finish after the function returns and its late RX result can be mistaken for the first frame of the next transfer. The abort path needs to quiesce the channel, or wait for and drain the in-flight frame, before reporting completion and updating the remaining count.

3. Pinmux policy should not be embedded in the reusable MCSPI LLD. `spi0_pinmux()` always remuxes CS0, CS1, and CS3, including `MCU_MCAN0_TX`, regardless of the selected channel. Starting SPI for channel 0 can therefore silently take CAN TX away. These selections should live in the board code or behind an explicit board hook.

4. The implementation does not honor the XHAL configuration-zero contract. `drvStart(&SPID1, NULL)` is defined to select configuration zero, but this driver returns `HAL_RET_CONFIG_ERROR`; similarly, `drvSelectCfgX(..., 0)` reapplies the current explicit configuration instead of selecting a stable default. Please provide a real configuration zero and use it consistently in both paths.

I would also validate `clock_mode <= 3` instead of silently masking higher values, and validate the initial configuration before resetting the controller (or fully roll back `ready` and the hardware state on failure).

The branch builds cleanly, but the reported hardware test exercises only the low-level polled path. Please add coverage for the public interrupt-driven transfer path, including normal completion followed by immediate unselect and an aborted transfer followed immediately by a new one.

gdisirio:
Disclosure: this review was prepared with assistance from OpenAI Codex.

arasnazmi:
Thanks, all four are fixed, and the coverage you asked for found a fifth problem in the interrupt-driven path itself: `MCSPI_TX` was written without waiting for `CHSTAT.TXS`, and a write to a full transmit register is dropped silently nothing shifts, no RX_FULL arrives, the frame count never moves. The polled path in this driver had always waited; every write to TX does now. A loop of plain interrupt-driven exchanges went from 6 of 8 stalling to 0 of 8. A single transfer always passed, which is why the earlier probe looked clean. 

1. **EOT.** `spi_lld_unselect()` and `spi_lld_stop()` wait for `CHSTAT.EOT` before releasing CS. EOT reads 0 out of reset, so it is only waited on once a frame has actually been sent (`shift_pending`).
2. **Abort.** `spi_lld_stop_transfer()` masks the interrupt, waits for the in-flight frame on `RXS`, collects it into the caller's buffer and counts it, then drains. Not on `EOT`: for a few cycles after a TX write it still reports the previous frame (`CHSTAT` reads `0x4` there on hardware).
3. **Pinmux.** Out of the LLD. `board_spi0_pinmux(channel)` muxes the shared signals plus the one chip select asked for, so channel 0 no longer takes `MCU_MCAN0_TX`; CS2 is refused.
4. **Configuration zero.** `SPI_DEFAULT_CONFIGURATION` from `AM67_SPI_DEFAULT_{SPEED,CLOCK_MODE,CS_CHANNEL}`, selected by `drvStart(spip, NULL)`, `spi_lld_setcfg(spip, NULL)` and `drvSelectCfgX(..., 0)` alike. `clock_mode > 3` is now refused rather than masked, and the initial configuration is validated before the reset.

The self-test adds the public paths: an interrupt-driven exchange followed immediately by `spiUnselectX()`, and a 16-frame burst aborted mid-flight with a new exchange right behind it, 8 rounds, all checked against the ICM-20948 `WHO_AM_I`. Waits are bounded, so a peripheral that stops answering fails the test instead of blocking the demo. Three power-on boots, 0xEA everywhere, 3/3.

One caveat: on the T3 Gemstone O1 image Linux binds the same instance and shares the interrupt, and its handler answers by zeroing the module-global `MCSPI_IRQENABLE`, which disarms a firmware transfer in progress. Validation
ran with the instance reserved for the firmware in the device tree; the self-test documents that requirement the way the SIO one does for the UART.

gdisirio (changes requested):
Thanks, the original EOT, abort, pinmux, validation, and test-coverage points are addressed well. The hardware findings are especially useful. I still see two issues to resolve before merging:

1. The new TX-ready timeout handling is not safe yet. In `spi_start_transfer()`, RX interrupts are enabled and the transfer fields are committed before waiting for `CHSTAT.TXS`. If that wait times out, the function returns failure but leaves the interrupt enabled; the generic layer has also already moved the driver to `HAL_DRV_STATE_ACTIVE`. In `spi_serve_interrupt()`, the return value from the same wait is discarded and TX is written anyway, recreating the silent dropped-write condition this change is intended to prevent. Please make the start failure path restore a clean READY/quiescent state, and make the ISR timeout disable the transfer and report an error rather than writing TX.

2. Configuration zero is still not the same configuration in all paths when `SPI_USE_CONFIGURATIONS == TRUE`: `drvStart(spip, NULL)` resolves to the built-in `spi_default_config`, while `drvSelectCfgX(spip, 0U)` resolves to `spi_configurations.cfgs[0]`. This contradicts the generic configuration-zero contract and the new comments saying those operations are equivalent. The same pattern exists in the RP and STM32 XHAL ports, so this may also expose a broader issue, but the new TI implementation should not document an equivalence it does not provide. Please use one resolver for configuration zero so both paths select the same object.

The updated branch builds cleanly, `git diff --check` passes, and the CI checks are green.

Disclosure: this follow-up review was prepared with assistance from OpenAI Codex.

arasnazmi:
Both fixed in 1b31319.

1. The TXS wait now happens first in `spi_start_transfer()`, before any driver field is set and before RX_FULL is enabled, so a timeout leaves the driver and the module as they were found. In `spi_serve_interrupt()` the result is no longer discarded: on timeout it masks the interrupt and reports the failure through `_spi_isr_error_code()` instead of writing TX, leaving the frames not exchanged in `remaining` for `spi_lld_stop_transfer()`.

One part of this is outside the LLD: `spiStartExchangeI()` sets `HAL_DRV_STATE_ACTIVE` before calling the LLD and does not revert it when the LLD returns an error, so the driver stays ACTIVE until the caller stops the transfer -- the same for the pre-existing `!ready` early returns. That lives in the generated `hal_spi.c`, so I have not touched it here; happy to send a template patch separately if you want it.

2. `spi_cfg_zero()` is now the single resolver: entry zero of the board's `spi_configurations` array when there is one, the built-in default otherwise. `drvStart(spip, NULL)`, `spi_lld_setcfg(spip, NULL)` and `drvSelectCfgX(spip, 0U)` all go through it, so the equivalence the comments claim holds in both configurations. I built and linked the `SPI_USE_CONFIGURATIONS == TRUE` path against a stub array to check it.

Re-validated on hardware, instance reserved for the firmware: three power-on boots, polled / interrupt-driven / abort-then-resume all 0xEA, 8 abort rounds per boot, 3 of 3.

PR #321, opened 2026-09-17
STM32U5: add retention-aware Stop clock resume

CInsights:
## Summary

- add an optional `halClockResumeMode()` operation for retained low-power clock configurations;
- implement a retention-aware Stop 0/1/2 resume path in the STM32U5 HAL; and
- mirror the API and implementation in HAL and XHAL.

After Stop, `halClockSwitchMode()` performs a deliberate whole-tree reconfiguration. On STM32U5 this temporarily rewrites the MSI ranges and peripheral clock selectors. That is appropriate for an arbitrary mode switch, but it can disturb an autonomous peripheral which is already active on a retained MSIK kernel clock during wakeup.

The new operation first establishes the live wake HCLK and validates both the requested tree and the registers which Stop must have retained. It then restores only state which can be lost or limited by Stop: FLASH latency, voltage range, required oscillators, the booster, the Stop-limited MSIS range, PLL enable state and `CFGR1.SW`. Retained MSIK range, PLL configuration, bus prescalers and CCIPR selectors are validated but not rewritten.

A mismatch or bounded-wait failure is returned to the caller. The caller can explicitly select the existing destructive mode switch as a recovery policy; the resume primitive does not silently fall back.

## Related

- Issue(s): none
- Notes for reviewers: the API is additive and enabled only when STM32U5 dynamic clock management is enabled. The caller must keep interrupts masked through the resume operation. The STM32U5 implementation accepts the documented MSIS or HSI16 Stop wake sources.

## Target Branch Check

      merged on `master`

## Scope

## Mechanical Checks

- [x] Style check passed on the changed code (the existing `//` findings later in `os/xhal/include/hal.h` are unchanged by this PR)

Upstream CI passed its POSIX simulator and ARM smoke builds. Additional affected-target builds passed locally:

- xPack GCC 15.2.1 / build-tools 4.4.1: custom STM32U5F7 classic HAL image with dynamic clocks enabled;
- xPack GCC 15.2.1 / build-tools 4.4.1: STM32U575 XHAL test image with dynamic clocks enabled.

Both local test projects were cleaned after the builds.

## Testing

The exact final patch was tested after a debugger-free cold boot on a custom STM32U5F7 board. The configuration used MSIS as the Stop wake clock, retained 24 MHz MSIK as the autonomous LPUART kernel clock, and restored HSE/PLL1R as the 160 MHz RUN clock.

In a 120-second 10 ms polling window the board completed 13,850 Stop resumes with zero resume failures, no UsageFault or HardFault, and 4,095 successful addressed polls out of 4,095. Register snapshots confirmed that the MSIK range and peripheral selector remained unchanged across resume.

## Compatibility and Risk

This is an additive API guarded by `HAL_LLD_USE_CLOCK_RESUME`; existing `halClockSwitchMode()` behavior is unchanged. The retained resume is intentionally strict: unexpected retained-register state is reported as failure rather than repaired in place. Applications which want recovery must explicitly invoke the general clock switch after handling that failure.


gdisirio (changes requested):
I am requesting an API change before merging this.

`halClockResumeMode(ccp)` cannot resume an arbitrary clock configuration: `ccp` must describe the configuration that was active immediately before entering Stop. The implementation validates the retained state against it and restores the parts that Stop altered. Passing any other configuration is therefore invalid, but the current signature makes this look like another mode-selection operation and leaves room for a caller/configuration mismatch.

I think the HAL should retain the current configuration and expose a parameterless operation, for example:

```c
bool halClockResume(void);
```

The LLD could keep a copy of the last successfully installed `halclkcfg_t`: initialize it after the default configuration is installed, and replace it only after a successful `halClockSwitchMode()`. Resume would then always restore that saved configuration. A copy is preferable to retaining the caller-provided pointer because the current switching API does not impose static lifetime on configuration objects.

This also makes the intended distinction explicit: resume restores the last active configuration; selecting a different one remains the job of `halClockSwitchMode()`.

Please also put the execution and failure contract on the public HAL/XHAL API: interrupts must remain masked throughout resume, and a failure may leave a partially restored clock state, requiring recovery before interrupts are unmasked.

This review was prepared with assistance from OpenAI Codex.

CInsights:
Agreed. `halClockResumeMode(ccp)` exposes more choice than the operation can validly support and makes a caller/configuration mismatch possible. I’ll revise the API to parameterless `halClockResume()`.

The STM32U5 LLD will retain a copy of the last successfully installed `halclkcfg_t`: initialize it after the default configuration is installed, update it only after a successful `halClockSwitchMode()`, and use that retained copy for resume. It will not retain the caller’s pointer.

I’ll also document on the public HAL/XHAL API that interrupts must remain masked throughout resume and that, on failure, the clock state may be only partially restored; the caller must recover it before unmasking interrupts.

I’ll update both HAL and XHAL implementations and re-run the custom STM32U5F7 hardware tests, including repeated STOP2 wake and the 10 ms LPUART polling workload, before pushing the revision. Thank you.

CInsights:
Implemented the requested API and ownership change in `2baeaef9bd`.

- The public operation is now parameterless `halClockResume()` in both HAL and XHAL.
- The STM32U5 LLD owns a copy of the active `halclkcfg_t`; it is initialized after the default configuration is active and replaced only after a successful `halClockSwitchMode()`.
- Resume always uses that retained copy, so no caller pointer or arbitrary configuration can be supplied.
- The public HAL/XHAL contract now states that interrupts must remain masked throughout resume and that failure may leave a partially restored clock state which must be recovered before unmasking.

Validation completed with xPack GCC 15.2.1:

- classic HAL dynamic-clock build: `demos/STM32/RT-STM32-XSHELL`, STM32U575 target;
- XHAL dynamic-clock build: `testxhal/ADC-GPT`, STM32U575 target;
- exact revised source integrated into the normal workload on a custom STM32U5F7 board, programmed/verified and then tested after a debugger-free cold boot. Qualification ELF SHA-256: `CB8AD143D714B4E890BDC3D4673443BC3E53C0FC85D6ECD2EBDADE9D13139F9B`.

The hardware window ran for 120.426 seconds with 10 ms controller scheduling, the U5F7 board entering genuine Stop 2, and two control nodes held in Run. The U5F7 board completed 7,107 retained-clock resumes with zero resume failures, no retained halt/UsageFault/HardFault, and the capture/convert/store workload remained healthy. The Run-held controls' resume counters did not change.

The controller recorded 9,307 poll starts: 9,303 successful and four transport timeouts, all on the Stop-2 board. Those four are the separately isolated receive-boundary issue where the start bit/marked master-mode address is lost; they occurred with clean clock-resume diagnostics and do not indicate a resume failure. All nodes remained online. I am keeping that investigation separate from this clock-resume API/fix.

gdisirio (approved):

